PR #127, opened 2022-06-14
Generics

jmendesky:
Relates to https://github.com/sky-uk/kfp-operator/pull/126

In addition to the above PR, this also removes duplicated test code by introducing generics


paoloambrosio-skyuk:
Closing as we have restarted this in a better way in #176.

PR #147, opened 2022-07-25
Downgrade all resource versions

### LINKED ISSUE #146: Downgrade resource versions to v1alpha1
# Overview

All resources have been created with version `pipelines.kubeflow.org/v1`, which suggests a stable version.
The project, however, is still in a pre-release state and the resources should reflect this.

**This is an unanticipated breaking change that is needed to allow consistent versioning in future releases.**

# AC

- [x] All resources downgraded to `pipelines.kubeflow.org/v1alpha1`
- [x] Examples and docs updated

jmendesky:
Closes #146 

## Tasks

- [x] Downgrade CRDs
- [x] Downgrade Config
- [x] Documentation updated


config/crd/bases/config.kubeflow.org_kfpcontrollerconfigs.yaml
```diff
@@ -16,7 +16,7 @@ spec:
     singular: kfpcontrollerconfig
   scope: Namespaced
   versions:
-  - name: v2
```
paoloambrosio-skyuk, line 19: Why was this v2?!

jmendesky: It was autogenerated by the CLI. Note that this is the spec for the controller config.
this here gives differing suggestions for what it should be: https://book.kubebuilder.io/component-config-tutorial/define-custom-config.html
IMO this should be v1alpha1 as well

paoloambrosio-skyuk (approved):

Issue #31, opened 2021-10-27
Support Beam arguments with multiple values


jmendesky:
## Overview

The `beamArgs` and `defaultBeamArgs` configuration options should support keys with multiple values because Beam itself supports them.

For example, Beam allows the specification of multiple experiments as follows: `--experiments="use_runner_v2" --experiments="shuffle_mode=service"`

Currently, this is not supported by either `beamArgs` or `defaultBeamArgs` fields:

```yaml
beamArgs:
  experiments: use_runner_v2
```

## Solution

```yaml
beamArgs:
- name: experiment
  value: exp1
- name: experiment
  value: exp2
```

### Notes:

- The resulting resource version would not change from previous behaviour
- `defaultBeamArgs` will be added to the final list of arguments instead of overriding them. A new solution for overriding args should be found
- For consistency and because of what we have found out in https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps, we are also going to change the `env` field in pipelines and `runtimeParameters` in runconfigurations
- This change will increase CRD versions to `v1alpha3`

## DIscarded solution 1: provide arguments as a list

```yaml
defaultBeamArgs:
- --project=gcp_project
- --temp_location=/tmp
```

```yaml
beamArgs:
- --experiment="use_runner_v2"
- --experiment="shuffle_mode=service"
```

### Notes
- Unintuitive syntax
- The resulting resource version would change from previous behaviour

## Discarded solution 2: allow single or multiple values

```yaml
defaultBeamArgs:
  project: gcp_project
  temp_location: /tmp
```

```yaml
beamArgs:
  project: gcp_project
  experiment:
  - use_runner_v2
  - shuffle_mode=service
```

### Notes
- This solution would be backwards compatible, but it violates the Kubernetes [API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps)
- - CRD generation of `map[string]interface{}` is not easily achievable or desirable https://github.com/kubernetes-sigs/controller-tools/pull/528.

## Discarded Solution 3

```yaml
beamArgs:
- name: project
  value: gcp_project
- name: experiment
  vaues:
  - use_runner_v2
  - experiments="shuffle_mode=service"
```

### Notes
- The resulting resource version would not change from previous behaviour but we would have to handle edge cases where both `Value` and `Values` could be set

PR #152, opened 2022-08-18
Add capability of compiler to parse beam args with list values

EdwardCuiPeacock:
Allowing the compiler to parse beam args with list values would allow multiple `--experiments` flags being used. This is related to issue https://github.com/sky-uk/kfp-operator/issues/31

@jmendesky @adriangay @alexgeorgousis


EdwardCuiPeacock:
@adriangay @jmendesky

PR #158, opened 2022-09-01
Helm installation from published packages

jmendesky:
- Default Helm values to point to the package repository
- Disabled manager monitoring and eventsource server by default
- Provide basic installation example
- Project root target to generate the website

PR #163, opened 2022-09-08
Fix integration tests by using local container images

jmendesky:
https://github.com/sky-uk/kfp-operator/pull/158 has changed the default `containerRegistry` value, resulting in the integration tests failing as they were trying to pull all images.


paoloambrosio-skyuk (approved):
Good catch! 👏

Issue #175, opened 2022-10-13
RunConfiguration runtime parameters not created


paoloambrosio-skyuk:
**Describe the bug**

Runtime parameters defined in RunConfiguration resources are not created by providers.

**To Reproduce**

Create a RunConfiguration with runtime parameters. The uploaded RunConfiguration on the provider will not have them.

**Expected behavior**

Runtime parameters should be present in the provider.

**Version and configuration**

Version 0.2.1, any configuration.

**Logs**

N/A

**Additional context**

Runtime parameters are not passed to the Argo Workflow for creates and updates.

PR #176, opened 2022-10-14
Experiment and runconfig rework

### LINKED ISSUE #170: Experiments managed via provider interface
Implement experiments in the provider interface. Vertex AI would fail until it's implemented, not part of this issue.

jmendesky:
Closes #170 

## Tasks

- [x] Rework Experiments
- [ ] Rework RunConfigurations
- [ ] Consolidate Experiment and RR WF factories


argo/providers/base/provider.go
```diff
@@ -15,8 +15,29 @@ type PipelineDefinition struct {
...
+	Description string `yaml:"description"`
+}
+
+type RunConfigurationDefinition struct {
+	Name            string `yaml:"name"`
+	Version         string `yaml:"version"`
+	PipelineName    string `yaml:"pipelineName"`
+	PipelineVersion string `yaml:"pipelineVersion"`
+	ExperimentName  string `yaml:"experimentName"`
+	Schedule        string `yaml:"schedule"`
```
paoloambrosio-skyuk, line 30: Shall we add the runtime parameters here even if they are not used by the KFP implementation (#175)?

argo/providers/base/provider.go
```diff
@@ -15,8 +15,29 @@ type PipelineDefinition struct {
...
+}
+
 type Provider[Config any] interface {
 	CreatePipeline(providerConfig Config, pipelineDefinition PipelineDefinition, pipelineFile string, ctx context.Context) (string, error)
 	UpdatePipeline(providerConfig Config, pipelineDefinition PipelineDefinition, id string, pipelineFile string, ctx context.Context) (string, error)
-	DeletePipeline(providerConfig Config, pipelineDefinition PipelineDefinition, id string, ctx context.Context) error
+	DeletePipeline(providerConfig Config, id string, ctx context.Context) error
+
+	CreateRunConfiguration(providerConfig Config, runConfigurationDefinition RunConfigurationDefinition, ctx context.Context) (string, error)
```
paoloambrosio-skyuk, line 38: Wondering if the parameter name should. be something like `resourceDefinition` or `resource` instead of `runConfigurationDefinition` (for pipeline and experiments as well). After all we use a generic `id` instead of `runConfigurationId`.

argo/providers/base/provider_app.go
```diff
@@ -9,17 +9,25 @@ import (
...
-	PipelineFileParameter:       "pipeline-file",
-	OutputParameter:             "out",
+	PipelineDefinitionParameter:         "pipeline-definition",
+	ExperimentDefinitionParameter:       "experiment-definition",
+	RunConfigurationDefinitionParameter: "runconfiguration-definition",
+	ProviderConfigParameter:             "provider-config",
+	PipelineIdParameter:                 "pipeline-id",
+	ExperimentIdParameter:               "experiment-id",
+	RunConfigurationIdParameter:         "runconfiguration-id",
+	PipelineFileParameter:               "pipeline-file",
```
paoloambrosio-skyuk, line 29: Guess these will be renamed to `resource-definition`, `resource-id`, `resource-file` later in this PR after refactoring. We should be able to create generic workflow steps for create/update with and without file and a single delete.

paoloambrosio-skyuk (commented):
Some reminders for later.

controllers/pipelines/experiment_state_handler.go
```diff
@@ -226,16 +226,16 @@ func (st ExperimentStateHandler) onUpdating(ctx context.Context, experiment *pip
 			return From(experiment.Status).WithSynchronizationState(apis.Failed).WithMessage(failureMessage)
 		}
 
-		idResult, _ := getWorkflowOutput(succeeded, ExperimentWorkflowConstants.ExperimentIdParameterName)
+		result, _ := getWorkflowOutput(succeeded, WorkflowConstants.ProviderOutputParameterName)
```
paoloambrosio-skyuk, line 229: If `providerError` is not empty it shoudl fail.

argo/providers/base/provider.go
jmendesky: let's do it when we address that card

controllers/pipelines/experiment_state_handler.go
jmendesky: this has now been addressed

argo/providers/base/provider_app.go
paoloambrosio-skyuk: Raised #178.

paoloambrosio-skyuk (approved):
We are terrible Go developers: removed 800 lines while adding functionalities! 🤯

paoloambrosio-skyuk (approved):

PR #177, opened 2022-10-15
Update .tool-versions

jmendesky:
- Include golang version
- Upgrade protoc to support M1 chips

This also changes some generated files due to the golang generators adding whitespace


jmendesky:
ok to test

paoloambrosio-skyuk (approved):

PR #180, opened 2022-10-28
Rename KfpId to ProviderId

jmendesky:
KfpId is no longer appropriate as we now support Vertex AI as well

## Tasks

- [x] Documentation updated


main.go
```diff
@@ -34,6 +34,7 @@ import (
 	config "github.com/sky-uk/kfp-operator/apis/config/v1alpha3"
 	pipelinesv1alpha2 "github.com/sky-uk/kfp-operator/apis/pipelines/v1alpha2"
 	pipelinesv1alpha3 "github.com/sky-uk/kfp-operator/apis/pipelines/v1alpha3"
+	pipelinesv1alpha4 "github.com/sky-uk/kfp-operator/apis/pipelines/v1alpha4"
```
paoloambrosio-skyuk, line 37: There are `pipelinesv1alpha3` unchanged references in this file. I think we should use `pipelinesv1` for the latest version so that we don't make mistakes.

jmendesky: this is now done

paoloambrosio-skyuk (approved):

PR #187, opened 2022-11-29
Support multiple providers

### LINKED ISSUE #171: Support multiple providers in a single instance
Allow multiple providers in a single instance of the operator.

- Provider configuration separate from the controller's configuration, in the workflow namespace
  - Mount provider configuration in workflows instead of passing it from the controller
  - One entry/file per provider in a single configmap `{name}`
  - Helm to duplicate provider-specific resources (e.g. VAI) for each provider - we will decide later how to better handle this. We need to use the right provider config for run enqueuer, run submitter and run completion event source. There might be providers wanting to use the same topics, and providers that don't.
- Provider name as a resource annotation or default in controller configuration
  - If not specified, the default will be set in the status if not already present
- For backward compatibility, the provider might not be set for a resource. In that case, it will be set the first time the resource has changed.

This issue makes the following assumption, which future work may address:
- Resource references (e.g. pipeline and experiment in the run configuration) are not checked. Incompatible provider configuration may result in the provider failing, but not the operator.
- Provider for a resource cannot be changed during its lifetime - the operation should fail if it does

paoloambrosio-skyuk:
Closes #171.

## Tasks

- [x] Multiple provider configuration 
  - Providers exposed as separate ConfigMap
  - ConfigMap loaded by Argo Workflow and not controller
  - Helm to generate resources for multiple providers
- [x] Default provider in controller
- [x] Annotations to define provider for a resource
- [x] Documentation updated


jmendesky:
ok to test

jmendesky (approved):
:shipit:

PR #190, opened 2022-12-14
Unify Workflows

### LINKED ISSUE #178: Unify Argo workflow interface
Currently we have one workflow for each resource type. After the refactoring in #176, we are in a position to:
- Create two generic workflows, for direct or pre-processed (compiled) resources, that can share most steps
  - Alter the Provider interface to accept `resource-{param}` (e.g. `resource-id`) instead of `{recourceType}-{param}`
- ~Converge state handlers for all resources into one (it should be possible now by making them generic)~
- Remove Workflow operation labels, that should not be relevant anymore
- ~Rewrite KFP Provider to use APIs instead of CLI 🎉~ (Done in #191)

This will allow to structure our testing in a more effective way:
- Unit tests of state handler (testing all possible Workflow combinations of success with id/providerError combinations and failure)
- Decoupled tests between controller runtime and Workflow resources
- Integration tests in Minikube of Workflows with stub Provider and Compiler (this tests the interface between controller and Workflow, Workflow to Compiler, Workflow to Provider CLI)
  - They can be run in parallel now, given that we won't rely on remote stubs
  - ~~Idea: make the stub Provider call an HTTP endpoint that returns the output YAML (this is required to verify that the CLI was called with the right parameters)~~ We ended up programming the stub with a ConfigMap
- Integration tests of Providers against a real service (this tests Provider CLI interactions with an actual service)
- Integrated tests with both Providers (happy path only)

Note:
- We should document the testing trategy

jmendesky:
Closes #178 

- Consolidate Experiment and RunConfiguration workflow template into a single `simple` workflow template.
- Move Pipeline workflow template into `compiled` workflow template.
- Move `delete` template into `common` as it doesn't change between simple and compiled.
- Remove operation labels as they are not needed anymore

TODO: 
- [x] Introduce simple & compiled resource workflows
- [x] Rewrk Workflow factories
- [x] Remove Workflow Operation Labels


argo/providers/stub/provider.go
```diff
@@ -0,0 +1,94 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"github.com/sky-uk/kfp-operator/providers/base"
+	"github.com/sky-uk/kfp-operator/providers/base/generic"
+)
+
+type StubProviderConfig struct {
+	ExpectedOutput base.Output `yaml:"expectedOutput"`
+	ExpectedId     string      `yaml:"expectedId"`
```
jmendesky, line 13: `ExpectedOutput` should be `StubbedOutput`
`ExpectedId` should move into a stuct like `ExpectedInput.Id`

controllers/pipelines/suite_integration_test.go
```diff
@@ -50,12 +57,102 @@ var _ = BeforeSuite(func() {
...
+		ExpectedOutput: base.Output{},
+		ExpectedId:     resource.GetStatus().ProviderId.Id,
+	})
+}
+
+func FailDeletion[R pipelinesv1.Resource](resource R) base.Output {
+	return StubProvider(stub.StubProviderConfig{
+		ExpectedOutput: base.Output{
+			ProviderError: "an error occurred",
+			Id:            resource.GetStatus().ProviderId.Id,
```
jmendesky, line 129: `StubbedOutput.Id` should change every time

controllers/pipelines/resource_workflow_factory.go
```diff
@@ -0,0 +1,172 @@
...
+
+type WorkflowFactory[R pipelinesv1.Resource] interface {
+	ConstructCreationWorkflow(provider string, resource R) (*argo.Workflow, error)
+	ConstructUpdateWorkflow(provider string, resource R) (*argo.Workflow, error)
+	ConstructDeletionWorkflow(provider string, resource R) (*argo.Workflow, error)
+}
+
+type ResourceWorkflowFactory[R pipelinesv1.Resource] struct {
+	Config            config.Configuration
+	Compiled          bool
```
jmendesky, line 50: Replace with `TemplateNameGenerator` with 2 defaults (simple & compiled)

paoloambrosio-skyuk (approved):

PR #196, opened 2023-01-20
Add Run resource

### LINKED ISSUE #64: Manage one-off pipeline runs declaratively
## Overview

We want to support managing Kubeflow Pipelines runs with a new Run resource.

### Technical notes

-  `RunSpec` should be immutable (i.e. re-submission will not be allowed)
- A TTL configuration similar to K8s Job's `ttlSecondsAfterFinished` should allow the run to be deleted after completion
- One-off runs should work for both fixed pipeline versions and latest pipeline versions

### Implementation Notes

The work can be broken up into the following steps
- [x] Run with fixed pipeline versions (ignoring updates) #196 
- [x] Implement a Validating Webhook to make the Spec immutable #202 
- [x] Introduce `CompletionState` to Run's `Status` field #207
  - Depends on #89
- [x] Implement TTL #209
- [x] Allow pipeline reference to be specified without version #210
- [x] Update docs #212

jmendesky:
Relates to #64

Supports only fixed pipeline versions and fails on updates (addressed in subsequent changes)


argo/providers/vai/provider.go
```diff
@@ -282,34 +294,48 @@ func (vaip VAIProvider) DeleteExperiment(_ context.Context, _ VAIProviderConfig,
...
+		labels.PipelineVersion: runIntent.PipelineVersion,
+	}
+
+	var runId string
+
+	if runIntent.RunConfigurationName != "" {
+		runId = fmt.Sprintf("rc-%s-%s", runIntent.RunConfigurationName, uuid.New().String())
+		runLabels[labels.RunConfiguration] = runIntent.RunConfigurationName
+	} else {
+		runId = fmt.Sprintf("rc-%s", uuid.New().String())
```
paoloambrosio-skyuk, line 317:
```suggestion
		runId = fmt.Sprintf("run-%s", uuid.New().String())
```

controllers/pipelines/run_controller_decoupled_test.go
```diff
@@ -0,0 +1,60 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"github.com/sky-uk/kfp-operator/apis"
+	pipelinesv1 "github.com/sky-uk/kfp-operator/apis/pipelines/v1alpha4"
+	providers "github.com/sky-uk/kfp-operator/providers/base"
+	v1 "k8s.io/api/core/v1"
+)
+
+var _ = Describe("Run controller k8s integration", Serial, func() {
+	When("Creating, updating and deleting", func() {
```
paoloambrosio-skyuk, line 17:
```suggestion
	When("Creating and deleting", func() {
```

paoloambrosio-skyuk (approved):

paoloambrosio-skyuk:
OK to test

PR #201, opened 2023-01-31
Change model location for VAI pipeline runs

jmendesky:
The URI is not the location of the pushed model, but the base for pipeline step artifacts (which is present for any pipeline run artifact).
This has been changed to use the `pushed_destination` property instead, which is inline with the KFP equivalent.

### DIFF
--- a/argo/providers/vai/eventing_server.go
+++ b/argo/providers/vai/eventing_server.go
@@ -13,9 +13,9 @@ import (
 )
 
 const (
-	PushedModelArtifactType     = "tfx.PushedModel"
-	ModelPushedMetadataProperty = "pushed"
-	ModelPushedMetadataValue    = 1
+	PushedModelArtifactType        = "tfx.PushedModel"
+	ModelPushedMetadataProperty    = "pushed"
+	ModelPushedMetadataValue       = 1
 	ModelPushedDestinationProperty = "pushed_destination"
 )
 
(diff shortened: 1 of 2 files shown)


argo/providers/vai/vai_eventing_server_unit_test.go
```diff
@@ -225,15 +224,7 @@ var _ = Context("VaiEventingServer", func() {
...
-									{
-										Uri:         "gs://some/where",
-										SchemaTitle: "tfx.PushedModel",
-										Metadata: &structpb.Struct{
-											Fields: map[string]*structpb.Value{
-												"pushed": structpb.NewNumberValue(1),
-											},
-										},
-									},
+									artifact("gs://some/where"),
```
paoloambrosio-skyuk, line 227: Why is this the only one using the `artifact` function?

argo/providers/vai/eventing_server.go
```diff
@@ -107,9 +108,13 @@ func modelServingArtifactsForJob(job *aiplatformpb.PipelineJob) []ServingModelAr
...
 				if artifact.SchemaTitle == PushedModelArtifactType {
-					pushedProperty, ok := artifact.Metadata.AsMap()[ModelPushedMetadataProperty]
-					if ok && pushedProperty.(float64) == ModelPushedMetadataValue {
-						servingModelArtifacts = append(servingModelArtifacts, ServingModelArtifact{Name: name, Location: artifact.GetUri()})
+					properties := artifact.Metadata.AsMap()
+					pushedProperty, hasPushed := properties[ModelPushedMetadataProperty]
+					if hasPushed && pushedProperty.(float64) == ModelPushedMetadataValue {
+						pushedDestinationProperty, hasPushedDestination := properties[ModelPushedDestinationProperty]
+						if hasPushedDestination {
+							servingModelArtifacts = append(servingModelArtifacts, ServingModelArtifact{Name: name, Location: pushedDestinationProperty.(string)})
```
jmendesky, line 116: we should check that the property is actually a string or else we risk a runtime panic. This hasn't been done for now because we are not doing it in line 113 either. I'd be happy to add it (for both cases)

paoloambrosio-skyuk: I was checking how to do that to leave a comment. I think we should.

jmendesky: this is now done

argo/providers/vai/vai_eventing_server_unit_test.go
jmendesky: this has now been restructured

paoloambrosio-skyuk (dismissed):

paoloambrosio-skyuk (approved):

PR #202, opened 2023-02-01
Add validation Webhook for Runs

jmendesky:
Reject changes to Run resources as this resources is considered immutable.

Relates to #64


controllers/pipelines/run_controller_decoupled_test.go
```diff
@@ -34,6 +34,10 @@ var _ = Describe("Run controller k8s integration", Serial, func() {
 				g.Expect(run.Status.ProviderId.Provider).To(Equal(testConfig.DefaultProvider))
 			})).Should(Succeed())
 
+			Expect(runHelper.Update(func(run *pipelinesv1.Run) {
+				run.Spec = pipelinesv1.RandomRunSpec()
+			})).NotTo(Succeed())
```
paoloambrosio-skyuk, line 39: Is it possible to verify that it's the expected error and not failing for a different reason?

helm/kfp-operator/templates/webhook_manifests.yaml
```diff
@@ -0,0 +1,33 @@
+
```
paoloambrosio-skyuk, line 1:
```suggestion
```

paoloambrosio-skyuk (approved):

PR #210, opened 2023-03-13
Allow runs without fixed pipeline version

alexgeorgousis:
Relates to #64 

Currently we assume a pipeline version will be specified on every `Run`. So in this PR we implement `Run` controller logic to fetch the dependent pipeline version for `Run`s that don't specify a version. We copied the `ObservedPipelineVersion` concept from `RunConfiguration`s to enable this.

## Tasks
- [x] Refactoring


apis/pipelines/v1alpha4/depending_on_pipeline_resource.go
```diff
@@ -0,0 +1,8 @@
+package v1alpha4
+
+// +kubebuilder:object:generate=false
+type DependingOnPipelineResource interface {
```
jmendesky, line 4: This is needed to allow handling dependencies from Runs and RunConfigurations to Pipelines in a generic way

jmendesky: In the future, this could become generic on the dependency type

controllers/pipelines/depending_on_pipeline_reconciler_test.go
alexgeorgousis, line 1: This test used to be in `runconfiguration_controller_test.go`

alexgeorgousis: We move the function under test, so we moved the whole test file

controllers/pipelines/experiment_controller.go
```diff
@@ -56,16 +50,17 @@ func (r *ExperimentReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	return ctrl.Result{}, nil
 }
 
-func (r *ExperimentReconciler) reconciliationRequestsWorkflow(workflow client.Object) []reconcile.Request {
-	return r.BaseReconciler.reconciliationRequestsWorkflow(workflow, &pipelinesv1.Experiment{})
+type ExperimentFoo struct{}
```
paoloambrosio-skyuk, line 53: Great name!

controllers/pipelines/pipeline_controller.go
```diff
@@ -62,16 +56,17 @@ func (r *PipelineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 	return ctrl.Result{}, nil
 }
 
-func (r *PipelineReconciler) reconciliationRequestsWorkflow(workflow client.Object) []reconcile.Request {
-	return r.BaseReconciler.reconciliationRequestsWorkflow(workflow, &pipelinesv1.Pipeline{})
+type PipelineFoo struct{}
```
paoloambrosio-skyuk, line 59: 😍

paoloambrosio-skyuk (approved):

paoloambrosio-skyuk (approved):

PR #211, opened 2023-03-16
Document how to compile locally

jmendesky:
FAQ section describing how to execute the compiler locally to debug issues with Python environment variables (e.g. https://github.com/dynaconf/dynaconf/issues/869) or library loading.


alexgeorgousis:
> _build the compiler image as well as your pipeline's image and inject the pipeline's configuration._

Inject which configuration where? (from the point of view of someone not familiar with the compilation process)

Later on you have an instruction to "create the compiler configuration" which probably corresponds to that step, but it's still not clear what you're doing exactly.

I think you can follow the instructions just fine, because you have listed the exact commands you need to run, but some clarity could make it easier to understand what exactly you're doing every step of the way.

alexgeorgousis (approved):
Perfect ✅

alexgeorgousis (approved):

PR #233, opened 2023-05-03
Implement Run conversion between v1alpha4 and v1alpha5

jmendesky:
Conversion for runs hadn't been implemented when v1alpha5 was introduced.


paoloambrosio-skyuk (approved):

jmendesky:
ok to test

PR #236, opened 2023-05-05
Preserve triggers in RC conversion

jmendesky:
Store excess triggers in conversion annotation instead of erroring.

Open questions:
- should the order be preserved?
- why haven't we done this
```
spec:
  triggers:
    schedules:
    - 1 2 3 4 5
    - 2 3 4 5 6
    onChange: true
```

TODO:
- v1alpha2
- v1alpha3


paoloambrosio-skyuk:
Will be done in a different PR. We plan to change the structure to one that can be better described in the CRD and not require custom validation:

```yaml
spec:
  triggers:
    schedules:
    - 1 2 3 4 5
    - 2 3 4 5 6
    onChange:
    - pipeline
    runConfigurations:
    - upcoming
    - functionality
```

PR #259, opened 2023-06-21
Populate Artifacts from RunCompletion events

### LINKED ISSUE #250: Populate RunConfiguration Artifacts from RunCompletionEvent
Set Artifacts in RunConfiguration State as follows:

```yaml
kind: RunConfiguration
spec:
  run:
    ...
  triggers:
    ...
  artifacts:
  - name: an-artifact
     ...
status:
  latestRuns:
    succeeded:
      ...
      artifacts:
      - name: an-artifact
        location: gs://some/where
```

Depends on #246 and #249

jmendesky:
Closes #250

Note: This also fixes decoupled tests not being run for the status completer do to non-compliance with Go test naming


controllers/pipelines/runconfiguration_controller.go
```diff
@@ -98,7 +99,8 @@ func (r *RunConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	newStatus.SynchronizationState = newSynchronizationState
 	newStatus.Provider = desiredProvider
 
-	if newStatus != runConfiguration.Status {
+
+	if !reflect.DeepEqual(newStatus, runConfiguration.Status) {
```
jmendesky, line 103: RunConfigurationStatus is no longer comparable by using `==`. 
Reference: https://pkg.go.dev/reflect#DeepEqual

alexgeorgousis (approved):
💯

PR #268, opened 2023-08-02
Make namespace metadata configurable

jmendesky:
Introduce Helm configuration to allow specifying namespace metadata.


paoloambrosio-skyuk (approved):

jmendesky:
ok to test

PR #269, opened 2023-08-03
Unflake tests

jmendesky:
- [x] don't fail on empty arrays
- [x] remove triggers from dependency RCs to speed up tests as schedules don't need to be synced
- [x] fetch latest state of dependency RCs in order to prevent conflicts
- [x] wait for RC to go into succeeded after deleting triggers
- [x] pass for 3 consecutive times, which we all know is the developer's equivalent of statistical significance 😂

### DIFF
--- a/argo/status-updater/suite_decoupled_test.go
+++ b/argo/status-updater/suite_decoupled_test.go
@@ -4,6 +4,7 @@ package status_updater
 
 import (
 	"context"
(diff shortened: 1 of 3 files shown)


jmendesky:
ok to test

jmendesky:
ok to test

paoloambrosio-skyuk (approved):

PR #286, opened 2023-09-11
Deprecate v1alpha2-v1alpha4 and introduce v1alpha6

### LINKED ISSUE #245: v1beta1 - Drop status.SynchronizationState field
Breaking changes that will require a version change to v1beta1:
- remove `status.SynchronizationState` as this is duplicated in `status.conditions.SynchronizationSucceeded` (work done in https://github.com/sky-uk/kfp-operator/issues/278)
   - need to check that the summary line can pick the status out of the conditions list, need to filter list on type: Synchronized and then output the reason attribute.
   - currently all checks are done on synchronizationState field so all uses will have to move
```
  conditions:
  - lastTransitionTime: "2024-09-12T09:07:11Z"
    message: ""
    observedGeneration: 1
    reason: **Succeeded**
    status: "True"
    type: **Synchronized**
```

Tasks:
- [x] Decommission `v1alpha2`-`v1alpha4` (include removing logic "For migration from v1alpha4", etc.)
- [ ] change all usages of status.SynchronizationState to use conditions entry
- [ ] version migration code needed

jmendesky:
Relates to #245 

Moving the SynchronizationState will be moved in a separate PR

## Tasks

- [x] Remove v1alpha2...v1alpha4
- [x] Introduce v1alpha6
- [x] Move Observed/TriggerdPipelineversion

## Notes:
As specs haven't changed in this version, common types are kept in the `v1alpha6` directory only and referenced in the old types. This greatly reduces the amount of duplicate code.


jmendesky:
ok to test

jmendesky:
ok to test

jmendesky:
ok to test

apis/pipelines/v1alpha3/experiment_types.go
```diff
@@ -1,70 +0,0 @@
...
-
-	return fmt.Sprintf("%x", hash)
-}
-
-//+kubebuilder:object:root=true
-//+kubebuilder:subresource:status
-//+kubebuilder:resource:shortName="mlexp"
-//+kubebuilder:printcolumn:name="KfpId",type="string",JSONPath=".status.kfpId"
-//+kubebuilder:printcolumn:name="SynchronizationState",type="string",JSONPath=".status.synchronizationState"
-//+kubebuilder:printcolumn:name="Version",type="string",JSONPath=".status.version"
```
paoloambrosio-skyuk, line 31: Why was this removed in v4+?

PR #291, opened 2023-09-22
Complete RCs with one-off runs

### LINKED ISSUE #290: Complete RunConfiguration and resolve Artifacts triggered by one-off runs
**Describe the bug**
RunConfiguration status should be updated for any provider run for it.
We are currently only updating the status if the run was triggered through a schedule.

Also, artifacts are currently not propagated to these one-off runs.

**To Reproduce**
Create a RunConfiguration with a non-schedule trigger such as

```
spec:
  ...
  triggers:
    onChange:
    - runSpec
```

A run triggered by this will not update the RunConfiguration after completion.

**Expected behavior**
The RC should be updated

**Additional context**
This is likely caused by not propagating the runconfiguration's name to the provider.

jmendesky:
Closes #290 

## Tasks

- [x] Propagate artifacts
- [x] Complete RC by propagating labels

### DIFF
--- a/controllers/pipelines/runconfiguration_controller.go
+++ b/controllers/pipelines/runconfiguration_controller.go
@@ -21,6 +21,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+var RunConfigurationConstants = struct {
+	RunConfigurationNameLabelKey string
+}{
+	RunConfigurationNameLabelKey: apis.Group + "/runconfiguration.name",
+}
+
 // RunConfigurationReconciler reconciles a RunConfiguration object
 type RunConfigurationReconciler struct {
 	DependingOnPipelineReconciler[*pipelinesv1.RunConfiguration]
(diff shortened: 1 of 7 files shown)


controllers/pipelines/runconfiguration_controller.go
```diff
@@ -324,19 +330,18 @@ func findOwnedRuns(ctx context.Context, cli client.Reader, runConfiguration *pip
 }
 
 func (r *RunConfigurationReconciler) constructRunForRunConfiguration(provider string, runConfiguration *pipelinesv1.RunConfiguration) (*pipelinesv1.Run, error) {
+	spec := runConfiguration.Spec.Run
```
paoloambrosio-skyuk, line 333: This would populate artifacts and runtime parameters (that would be filled possibly from dependencies by the run controller)... correct?

jmendesky: runtime parameters in one-off runs are handled the same as in RunConfigurations (Resolution in the respective controller). We only pre-resolve them for RunSchedules at the moment. This change doesn't change this behaviour to what we had [before](https://github.com/sky-uk/kfp-operator/pull/291/files#diff-0a4eb9edb6dea35ffe018762cb542cec35b15b47e9151a04885cb5d7fc382a54L337).
So yes, this would be filled by the run controller

jmendesky:
ok to test

jmendesky:
ok to test

paoloambrosio-skyuk (approved):

PR #292, opened 2023-09-25
Succeed RC when dependencies are not met

### LINKED ISSUE #273: RC does not go into Succeeded when it fails to fetch dependent RC
Creating a RC that reference another RC that does not exist or it hasn't produced that artifect, leaves `status.synchronizationState` and `status.conditions` empty in the resource.

jmendesky:
Schedules and Runs should only be created once all dependencies have been met.

This also stabilizes some previously flakey tests

Closes #273

### DIFF
--- a/controllers/pipelines/runconfiguration_controller_decoupled_test.go
+++ b/controllers/pipelines/runconfiguration_controller_decoupled_test.go
@@ -252,9 +252,8 @@ var _ = Describe("RunConfiguration controller k8s integration", Serial, func() {
 
 			Expect(k8sClient.Status().Update(ctx, runConfiguration)).To(Succeed())
 
-			oldState := runConfiguration.Status.SynchronizationState
 			Eventually(matchRunConfiguration(runConfiguration, func(g Gomega, fetchedRc *pipelinesv1.RunConfiguration) {
-				g.Expect(fetchedRc.Status.SynchronizationState).To(Equal(oldState))
+				g.Expect(fetchedRc.Status.ObservedGeneration).To(Equal(fetchedRc.Generation))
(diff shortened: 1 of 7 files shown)


jmendesky:
ok to test

jmendesky:
ok to test

jmendesky:
ok to test

jmendesky:
ok to test

jmendesky:
ok to test

controllers/pipelines/runconfiguration_controller_decoupled_test.go
```diff
@@ -232,9 +252,8 @@ var _ = Describe("RunConfiguration controller k8s integration", Serial, func() {
 
 			Expect(k8sClient.Status().Update(ctx, runConfiguration)).To(Succeed())
 
-			oldState := runConfiguration.Status.SynchronizationState
 			Eventually(matchRunConfiguration(runConfiguration, func(g Gomega, fetchedRc *pipelinesv1.RunConfiguration) {
-				g.Expect(fetchedRc.Status.SynchronizationState).To(Equal(oldState))
+				g.Expect(fetchedRc.Status.ObservedGeneration).To(Equal(fetchedRc.Generation))
```
jmendesky, line 256: Because the state can now change to succeeded independent of dependencies, we can no longer assert on it

controllers/pipelines/suite_decoupled_test.go
paoloambrosio-skyuk, line 1: I would usually raise fixing flakey tests as a separate PR, but it's small and isolated enough so I think we are OK.

paoloambrosio-skyuk (approved):

PR #293, opened 2023-09-26
Remove additional level of JSON deserialisation

jmendesky:
In order not to have to further serialize on the consumer side, this PR removes an unnecessary level of deserialization.

Further testing is needed to assess this approach.


jmendesky:
Following this approach would result consumer events not to be in JSON format. Closing.

PR #294, opened 2023-10-06
Fix table in installation docs

paoloambrosio-skyuk:
There is a blank line in the middle of a Markdown table that [breaks its rendering](https://sky-uk.github.io/kfp-operator/docs/getting-started/installation/).


grahamia:
This was fixed in https://github.com/sky-uk/kfp-operator/pull/322

PR #296, opened 2023-10-26
Refactor argo-events gRPC API dependency

jmendesky:
References argo-events as a Golang dependency instead of downloading and compiling proto files.

this also required the following:
- tidy go dependencies
- re-generate mocks as dependencies have changed

[ml-metadata](https://github.com/google/ml-metadata) could not be refactored using this method as no Go packages are published.


grahamia (approved):

argo/providers/go.mod
```diff
@@ -3,96 +3,152 @@ module github.com/sky-uk/kfp-operator/argo/providers
...
 require (
-	cloud.google.com/go/aiplatform v1.26.0
-	cloud.google.com/go/pubsub v1.9.0
-	cloud.google.com/go/scheduler v1.7.0
-	cloud.google.com/go/storage v1.27.0
+	cloud.google.com/go/aiplatform v1.48.0
+	cloud.google.com/go/pubsub v1.33.0
+	cloud.google.com/go/scheduler v1.10.1
+	cloud.google.com/go/storage v1.30.1
+	github.com/argoproj/argo-events v1.8.1
```
paoloambrosio-skyuk, line 10: I imagine this is the line that really changed. I wish Go didn't have such an awful dependency management system.

argo/providers/vai/vai_eventing_server_unit_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build unit
+
```
paoloambrosio-skyuk, line 2: ...or build system 🤦‍♂️

argo/providers/go.mod
jmendesky: that is correct

argo/providers/vai/vai_eventing_server_unit_test.go
jmendesky: we found this when running `go vet` again on this directory

paoloambrosio-skyuk (approved):
As per usual, Go makes everything so much harder than it should be. Can't really assess if nothing else has changed, but it looks OK otherwise.

PR #297, opened 2023-11-02
Support TFX 1.14

adriangay:
This PR updates the KFP compiler build and test to support TFX 1.14. TFX 1.14 official container is built with Python 3.10, so compiler build/test is updated for that. It also removes support for Python 3.7, which is no longer required.

The following changes were required:
- dependencies updated
  - `python = ">=3.8,<3.11"` - removes support for 3.7; adds support for 3.10
  - `PyYAML = ">=5.4.1"` - this was updated in conjunction with a breaking change (see below)
  - `pytest = "7.4.0"` - this was found to be necessary to get some pipeline unit tests to work, so also upgraded here
  -  `shapely = "<2"` - see this [Google AI Platform issue](https://github.com/googleapis/python-aiplatform/issues/1852)
- updates to compiler `Dockerfile` to remove Python 3.7 and add 3.10
- fixed breaking change in compiler unit test - `yaml.load()` now needs a `Loader` arg, or use `safe_load()` method. The latter was used.
- updated `integration-test` Python versions
- fixed `integration-test` - was missing `--execution_mode` arg and compiler expects `--pipeline_config` arg to be a file
- in `integration-test` `Dockerfile`, the initial `shapely` downgrade on single `pip` command  `RUN pip install tfx==$TFX_VERSION "shapely<2"` did not work and introduced null errors and segmentation faults when importing TFX packages in `compiler.py`. Following the exact instructions in the issue above, ie. `pip install -U google-cloud-aiplatform "shapely<2"` fixed the issue
- tested `v1` and `v2` args on `--execution_mode`
-  while debugging the `shapely` issue, noticed `:://` in `PYTHONPATH` in `compile.sh`- it obviously did 'work' before, but corrected it anyway
- corrected compiler path in `README.md`

:white_check_mark: unit tests passing

:white_check_mark: acceptance test passing

:white_check_mark: integration test passing

:white_check_mark:  prBuild passing

Note: the issue with Python version during compile build/test was due to Poetry installation picking up Python 3.11.2. It does not 'pick' the 'currently active' version in a multi-Pyhon installation. There is no easy way for force it, but you can tell it which to use. The line:
```
poetry env use /home/jenkins/.asdf/installs/python/3.10.12/bin/python
```
in the compiler makefile forces Poetry to use a specified available version, in this case 3.10.12. the Python `bin` path in Jenkins is not the same as local machines, which is not ideal.


adriangay:
OK to test

adriangay:
OK to test

jmendesky:
ok to test

adriangay:
OK to test

argo/kfp-compiler/Makefile
```diff
@@ -13,22 +13,27 @@ penguin-staging:
...
+	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.10 --output_file /tmp/pipeline.yaml --pipeline_config /shared/pipeline_config.json  --execution_mode v1
+	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.9 --output_file /tmp/pipeline.yaml --pipeline_config /shared/pipeline_config.json --execution_mode v2
+	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.10 --output_file /tmp/pipeline.yaml --pipeline_config /shared/pipeline_config.json  --execution_mode v2
 
 ##@ Build
 
 build:
+	poetry env use /home/jenkins/.asdf/installs/python/3.10.12/bin/python
 	pip install poetry-dynamic-versioning --quiet
+	poetry config virtualenvs.create false
```
paoloambrosio-skyuk, line 36: This doesn't seem to be either necessary or a good idea. It changes the config for the user (in my case it changed `~/Library/Preferences/pypoetry/config.toml` that is never what we want). Also we don't want to install anything in the system packages:
> If set to false, Poetry will not create a new virtual environment. If it detects an already enabled virtual environment or an existing one in `{cache-dir}/virtualenvs` or `{project-dir}/.venv` it will install dependencies into them, otherwise it will install dependencies into the systems python environment.

argo/kfp-compiler/Makefile
```diff
@@ -13,22 +13,27 @@ penguin-staging:
...
-	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.9 --output_file /tmp/pipeline.yaml --pipeline_config '{"rootLocation": "pipeline_root", "servingLocation": "serving", "name": "test", "image": "test-pipeline", "tfxComponents": "pipeline.create_components"}'
+	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.9 --output_file /tmp/pipeline.yaml --pipeline_config /shared/pipeline_config.json --execution_mode v1
+	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.10 --output_file /tmp/pipeline.yaml --pipeline_config /shared/pipeline_config.json  --execution_mode v1
+	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.9 --output_file /tmp/pipeline.yaml --pipeline_config /shared/pipeline_config.json --execution_mode v2
+	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.10 --output_file /tmp/pipeline.yaml --pipeline_config /shared/pipeline_config.json  --execution_mode v2
 
 ##@ Build
 
 build:
+	poetry env use /home/jenkins/.asdf/installs/python/3.10.12/bin/python
```
paoloambrosio-skyuk, line 34: Of course this is incorrect as it will never build on any local machine.

argo/kfp-compiler/Makefile
adriangay: fixed

argo/kfp-compiler/Makefile
adriangay: fixed

argo/kfp-compiler/poetry.lock
```diff
@@ -1,100 +1,186 @@
+# This file is automatically @generated by Poetry 1.7.1 and should not be changed by hand.
```
paoloambrosio-skyuk, line 1: Poetry 1.7.1? It feels odd since we use 1.5.1... but I imagine it's backwards compatible?

paoloambrosio-skyuk (commented):
My main concern is how to document the shapely pinning to version <2. Should it be an issue on this repo that we link from both places with a "FIXME" comment?

argo/kfp-compiler/poetry.lock
adriangay: I had to reinstall Poetry. I recreated the lock file. I forgot abt this, but yes, compstible. Can change if you prefer, if I can get 1.5.1 back and working locally.

adriangay:
> My main concern is how to document the shapely pinning to version <2. Should it be an issue on this repo that we link from both places with a "FIXME" comment?

It should be noted somehow. At least referring to the issue it fixes would explain to OSS users why it's there.

.metals/metals.mv.db
jmendesky, line 1: This should not be here

argo/kfp-compiler/Makefile
```diff
@@ -13,17 +13,20 @@ penguin-staging:
 	cp ../../docs-gen/includes/quickstart/penguin_pipeline/*.py $(PENGUIN_STAGING)
 	cp integration/Dockerfile $(PENGUIN_STAGING)
 
-penguin-37: penguin-staging
-	docker build --build-arg PYTHON_VERSION=3.7.12 --build-arg TFX_VERSION=1.9.1 -t penguin:3.7 $(PENGUIN_STAGING)
+penguin-310: penguin-staging
+	docker build --build-arg PYTHON_VERSION=3.10.12 --build-arg TFX_VERSION=1.14.0 -t penguin:3.10 $(PENGUIN_STAGING)
```
jmendesky, line 17: it would be better if this came after 3.9

argo/kfp-compiler/Makefile
```diff
@@ -13,17 +13,20 @@ penguin-staging:
...
-	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.9 --output_file /tmp/pipeline.yaml --pipeline_config '{"rootLocation": "pipeline_root", "servingLocation": "serving", "name": "test", "image": "test-pipeline", "tfxComponents": "pipeline.create_components"}'
+	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.9 --output_file /tmp/pipeline.yaml --pipeline_config /shared/pipeline_config.json --execution_mode v1
+	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.10 --output_file /tmp/pipeline.yaml --pipeline_config /shared/pipeline_config.json  --execution_mode v1
+	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.9 --output_file /tmp/pipeline.yaml --pipeline_config /shared/pipeline_config.json --execution_mode v2
+	docker run -v $(TMP):/shared --entrypoint /shared/compile.sh penguin:3.10 --output_file /tmp/pipeline.yaml --pipeline_config /shared/pipeline_config.json  --execution_mode v2
```
jmendesky, line 29: Probably not something we can solve right now but in the future, this would benefit from defining a compatibility matrix and then iterating over it instead of duplicating this line

argo/kfp-compiler/integration/Dockerfile
```diff
@@ -5,6 +5,7 @@ FROM python:$PYTHON_VERSION
 ARG TFX_VERSION
 
 RUN pip install tfx==$TFX_VERSION
+RUN pip install -U google-cloud-aiplatform "shapely<2"
```
jmendesky, line 8: if the shapely dependency doesn't work on master already, we should raise a separate PR for this

.gitignore
```diff
@@ -17,6 +17,8 @@ dist/
 
 # VS Code
 /.vscode/
+.bloop/
+project/metals.sbt
```
jmendesky, line 21: This should be here either as this is not a scala project

.metals/metals.mv.db
jmendesky: This should not be in the `.gitignore` either as this is not a scala project

argo/kfp-compiler/integration/Dockerfile
adriangay: If I do that, then that PR will have to be merged first, otherwise this PR will fail?

argo/kfp-compiler/resources/compile.sh
```diff
@@ -2,6 +2,6 @@
 
 PYTHON_VERSION=$(python3 -V | grep -Eo '[0-9]+\.[0-9]+')
 
-export PYTHONPATH="$PYTHONPATH:/$(dirname $0)/py$PYTHON_VERSION"
+export PYTHONPATH="$PYTHONPATH$(dirname $0)/py$PYTHON_VERSION"
```
jmendesky, line 5: When concatenating paths, we should always separate with an additional `:` character 
```suggestion
export PYTHONPATH="$PYTHONPATH:$(dirname $0)/py$PYTHON_VERSION"
```

argo/kfp-compiler/integration/Dockerfile
jmendesky: yes, that's what we should do

argo/kfp-compiler/resources/compile.sh
adriangay: noticed `:://` in actual PYTHONPATH when running - it obviously did 'work' before, but corrected it anyway. maybe not the best fix.

jmendesky: sure, but that was because the original `PYTHONPATH` ended in a `:` already - which is not the case usually. Most likely, an empty string got concatenated before. Best practice for paths is always to concatenate with a new `:` so that we don't have to make assumptions about how it was set before.

adriangay: ok

argo/kfp-compiler/Makefile
adriangay: i assume from the POV of OSS users' compatibility

argo/kfp-compiler/integration/Dockerfile
adriangay: ok, will do that later this week

.tool-versions
```diff
@@ -1,5 +1,5 @@
-python 3.7.12
-poetry 1.2.0
+python 3.10.12
+poetry 1.5.1
```
paoloambrosio-skyuk, line 2: Is there a reason why we are not upgrading directly to the latest 1.7.1?

argo/kfp-compiler/poetry.lock
paoloambrosio-skyuk: Maybe we should just upgrade to Poetry 1.7.1 in `.tool-versions`?

paoloambrosio-skyuk (commented):
Some more thoughts on versions:
- `.tool-versions`
  - :white_check_mark: We should always use the latest Python version from `pyproject.toml`
  - :negative_squared_cross_mark: We should always use the latest Poetry
- `argo/kfp-compiler/pyproject.toml`
  - :white_check_mark: we should pin tfx to the latest supported version
  - :white_check_mark: Python version range should be set [accordingly](https://github.com/tensorflow/tfx#compatible-versions)
- `argo/kfp-compiler/Makefile` and `argo/kfp-compiler/Dockerfile`
  - :negative_squared_cross_mark: We should build and test with all versions. In the PR 3.9 and 3.10 seem completely arbitrary, and 3.8 is missed.

argo/kfp-compiler/integration/Dockerfile
paoloambrosio-skyuk: After some digging, I believe that it should be part of this PR.

`shapely` is only a dependency of `google-cloud-aiplatform` of which:
- the latest and not working is `1.38.1`
- `tfx` `1.14` requires `>=1.6.2,<2`
- `apache-beam` requires `>=1.26.0,<2.0` (and this is a dependency of `tfx` BTW)

Any version `>=1.26.0,<1.35.0` would work. This [changed in `tfx` `1.14`](https://github.com/tensorflow/tfx/blob/v1.14.0/RELEASE.md#dependency-updates), that relaxed the version constraints on those dependencies.

.tool-versions
adriangay: no reason. i did upgrade locally myself just before xmas

argo/kfp-compiler/Makefile
jmendesky: given that we test versions in the acceptance tests, we can just test one version here (v2)

adriangay:
OK to test

.tool-versions
adriangay: done

argo/kfp-compiler/resources/compile.sh
adriangay: done

argo/kfp-compiler/integration/Dockerfile
adriangay: leaving as-is

argo/kfp-compiler/poetry.lock
adriangay: done

adriangay:
OK to test

paoloambrosio-skyuk:
OK to test

paoloambrosio-skyuk:
OK to test

paoloambrosio-skyuk:
OK to test

jmendesky (approved):

PR #299, opened 2023-11-30
RunCompletionEvent String function

grahamia:
Adding a String() function to RunCompletionEvent so can log without pointer references


jmendesky:
ok to test

PR #303, opened 2023-12-13
Add FlatMap and FlatMapErr

jmendesky:
Adds `FlatMap` and `FlatMapErr` functions for collection utils as shorthands for `Flatten(Map)`


grahamia (approved):
🥇

PR #306, opened 2024-01-10
Handle fullstops in pipeline version

### LINKED ISSUE #287: Pipeline versions cannot be run on VAI
**Describe the bug**
Pipeline versions are persisted onto VAI runs as a label. Because VAI labels only support lowercase letters, numbers, dashes and underscores some valid pipeline versions result in VAI refusing the creation of the job.

**To Reproduce**
Create a Pipeline with a version containing dots, for example

```yaml
apiVersion: pipelines.kubeflow.org/v1alpha5
kind: Pipeline
metadata:
  annotations:
    pipelines.kubeflow.org/provider: vai
  name: kfp-operator-quickstart
spec:
  image: ghcr.io/kfp-operator/kfp-operator-quickstart:0.4.0
  tfxComponents: pipeline.create_components
```

grahamia:
Parse the PipelineVersion to turn fullstops into dashes (handle semvar tagged build) so submit is  successful.

Only change for the label all other usage should stay the same.

Co-authored-by: Jonas Mende <jonas.mende@sky.uk>

Closes #287

### DIFF
--- a/argo/providers/vai/provider.go
+++ b/argo/providers/vai/provider.go
@@ -327,19 +327,11 @@ func generateRunId(runIntent RunIntent) string {
 	return fmt.Sprintf("%s", uuid.New().String())
 }
 
-func (vaip VAIProvider) EnqueueRun(ctx context.Context, providerConfig VAIProviderConfig, runIntent RunIntent) (string, error) {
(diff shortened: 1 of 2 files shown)


jmendesky (approved):

jmendesky:
ok to test

paoloambrosio-skyuk (approved):

PR #310, opened 2024-01-22
Provide provider name in RunCompletionEvents

### LINKED ISSUE #304: Add `provider` field to RunCompletionEvents
### Overview

Add a `provider` field to [RunCompletionEvent](https://sky-uk.github.io/kfp-operator/docs/reference/run-completion/).
This can be useful for consumers of these events to distinguish their source.

jmendesky:
Closes #304

### Overview:

Adds the `provider` field to RunCompletionEvents.
Each provider is responsible for propagating the configuration of this field to their respective backends.

### Alternatives considered:

Alternatively, the provider name could have been set as a label/in the description of all resources created on provider backends. Because currently there needs to be a one-to-one relationship between providers and backends, this information would be redundant. In the future, we may decide to add this.


jmendesky:
ok to test

grahamia (approved):

PR #312, opened 2024-01-30
Fix intermittent decoupled test failures

jmendesky:
## Overview

Fix intermittent decoupled test failures by waiting for state changes before attempting resource updates. We have introduced `runconfiguration_test_helper.go` in order to separate test utilities from actual tests.

Also address `aiplatform` import deprecation warnings.


grahamia (approved):
:shipit:

controllers/pipelines/run_controller_decoupled_test.go
```diff
@@ -261,9 +261,7 @@ var _ = Describe("Run controller k8s integration", Serial, func() {
 			run.SetDependencyRuns(map[string]pipelinesv1.RunReference{excessDependency: {}})
 			Expect(k8sClient.Status().Update(ctx, run)).To(Succeed())
 
-			oldState := run.Status.SynchronizationState
```
jmendesky, line 264: This would always succeed as `oldState` will be based on the `Update`

PR #322, opened 2024-03-08
Remove VAI provider cloud scheduler support

### LINKED ISSUE #285: Remove Support for Cloud Scheduler
## Overview
Now https://github.com/sky-uk/kfp-operator/issues/284 has been complete the next step is to remove the legacy cloud scheduler code along with the code that was introduced to migrate existing schedules over to native Vertex AI scheduler.

## Notes
- all code should be removed for the cloud scheduler pipeline
- all migration code should be removed
- any configuration/deployment setup for run-intents etc should be removed
- all documentation updated
- any diagram updated with simplified setup

grahamia:
Closes #285 

Following migration from cloud scheduler into native Vertex AI scheduler we can now remove the legacy scheduler code and migration code.

## Tasks:
- [X] update documentation
- [X] update diagrams
- [X] remove cloud scheduler code
- [X] remove migration code


argo/providers/vai/provider.go
```diff
@@ -36,8 +33,9 @@ var labels = struct {
 	RunConfigurationNamespace string
 	RunName                   string
 	RunNamespace              string
-	LegacyNamespace           string
-	LegacyRunConfiguration    string
+	// TODO SHOULD WE DELETE THESE NOW?
```
grahamia, line 36: @jmendesky

grahamia: comment in code is
// For compatability with resources created with v0.3.0 and older

jmendesky: This can be removed now.

argo/providers/vai/provider.go
```diff
@@ -228,7 +217,47 @@ func (vaip VAIProvider) CreateRun(ctx context.Context, providerConfig VAIProvide
...
+	for name, value := range run.RuntimeParameters {
+		parameters[name] = &aiplatformpb.Value{
+			Value: &aiplatformpb.Value_StringValue{
+				StringValue: value,
+			},
+		}
+	}
+
+	pipelineJob := &aiplatformpb.PipelineJob{
+		Labels:         run.Labels,
```
jmendesky, line 236: I think the `VAIRun` created above is not necessary. I fact, I think we can remove that struct entirely.

grahamia: Agreed and have removed, only 2 values used from struct so passing as params now

jmendesky (approved):
:shipit:

Issue #40, opened 2021-11-01
Support namespaced custom resources on providers


jmendesky:
### Overview

The KFP-Operator has been built without multi-user support. Custom resources, currently, are assumed to have globally unique names.
All providers should support duplicate names across namespaces.

#### Tasks

- all labeled resources should contain both the resource's name and namespace
- resource names should be namespaced

#### Open Questions

- can VAI Pipeline schedules have labels?

#### Notes

Migration will happen eventually as resources are synchronised. No manual migration procedures are needed.

#### Open Questions

KFP support requires research into mult-user isolation
https://www.kubeflow.org/docs/components/central-dash/profiles/

- install KFP with multi-user isolation, investigate and design a solution

aidandunlop:
Rather than using profiles in KFP, I think namespaced pipelines is already a feature? https://github.com/kubeflow/pipelines/pull/4835

PR #326, opened 2024-05-13
VAI Provider - Namespaced custom resources for pipelines (scheduled and one off)

grahamia:
Connects to #40.

Compiler sets pipeline_name which ends up being the pipelineInfo/name in the pipeline definition which is then used in both one off runs and scheduled runs to set the run name.

### Workflow Changes
RunScheduleDefinition and PipelineDefinition Name attribute changed from string (i.e. penguin-pipeline-name) to NamespacedName (i.e. namespace/penguin-pipeline-name) so these values are passed into the relevant workflow.

### Compile Changes
Changes were needed here because it was found there is currently no way to override the "scheduled run name" (i.e. the name given to individual runs) it just uses the pipelineInfo/name value. A request has been made to Google change the [createSchedule](https://pkg.go.dev/cloud.google.com/go/aiplatform@v1.67.0/apiv1/aiplatformpb#CreateScheduleRequest) interface in the Vertex AI Pipeline API to allow setting at creation time but until that is made available this has to be done.

As `/` characters are not valid as a pipeline name we sanitise it to a `-` and set that into the pipeline_name which then gets set into the pipelineInfo/name element on the pipeline yaml. pipeline_paths_for_config automatically includes the namespace in all paths. For example in the pipeline yaml we now have:
```
"base_directory": "gs://your-servicing-bucket/namespace/penguin-pipeline-name/serving"
```

### VAI Provider Changes
CreateRun changed to calculate PipelineJobId from pipelineInfo/name within the job spec (which had previously been processed in the compiler above). Changed this to use the pipeline spec so that both one off and scheduled runs are driven from same place.

GCS paths all changed to use namespaced names

Creating updating run schedule using passed in namespaced name from workflow to add namespace into the display name of the schedule in VAI. The "Scheduled run name" in VAI is picked up from the pipelineInfo/name field in the pipeline spec.

### KFP Provider Changes
Changed so functions as existing, name is used from within the passed NamespacedName value.

### Noticeable changes with VAI
* pipeline yaml uploaded into GCS within namespaced "directory"
* Scheduled run name includes namespace
* schedule display name includes namespace
* one off run named "namespace-penguin-pipeline-name-<random chars>"
* scheduled runs named "namespace-penguin-pipeline-name-<timestamp>"
* artifacts are pushed to namespaced directory in GCS

## Tasks

- [X] add namespace to one off runs
- [X] add namespace to schedules and scheduled run name
- [X] add namespace to storage path
- [X] Documentation updated

### DIFF
--- a/argo/kfp-compiler/acceptance/test_compiler.py
+++ b/argo/kfp-compiler/acceptance/test_compiler.py
@@ -8,7 +8,8 @@ from tempfile import TemporaryDirectory
 import pytest
 
 runner = CliRunner()
-config_file_path = 'acceptance/pipeline_conf.yaml'
+unnamespaced_config_file_path = 'acceptance/pipeline_conf.yaml'
+namespaced_config_file_path = 'acceptance/namespaced_pipeline_conf.yaml'
 
 
 def provider_config_file_path(execution_mode):
@@ -24,7 +25,7 @@ def test_cli_v1():
     with TemporaryDirectory() as tmp_dir:
         output_file_path = os.path.join(tmp_dir, 'pipeline.yaml')
 
-        result = runner.invoke(compiler.compile, ['--pipeline_config', config_file_path, '--provider_config', provider_config_file_path('v1'), '--output_file', output_file_path])
+        result = runner.invoke(compiler.compile, ['--pipeline_config', unnamespaced_config_file_path, '--provider_config', provider_config_file_path('v1'), '--output_file', output_file_path])
 
         assert result.exit_code == 0
         assert os.stat(output_file_path).st_size != 0
@@ -34,7 +35,11 @@ def test_cli_v1():
         assert workflow['kind'] == 'Workflow'
 
 
-def test_cli_v2():
+@pytest.mark.parametrize("config_file_path, expected_pipelineinfo_name", [
+    (unnamespaced_config_file_path,  "test"),
+    (namespaced_config_file_path, "namespace-test")
+])
+def test_cli_v2(config_file_path: str, expected_pipelineinfo_name: str):
     with TemporaryDirectory() as tmp_dir:
         output_file_path = os.path.join(tmp_dir, 'pipeline.yaml')
 
@@ -46,6 +51,7 @@ def test_cli_v2():
         f = open(output_file_path, "r")
         pipeline = yaml.safe_load(f.read())
         assert pipeline['pipelineSpec']['schemaVersion'] == '2.0.0'
+        assert pipeline['pipelineSpec']['pipelineInfo']['name'] == expected_pipelineinfo_name
 
 
 def test_failure():
--- a/argo/kfp-compiler/acceptance/test_compiler.py
+++ b/argo/kfp-compiler/acceptance/test_compiler.py
@@ -8,8 +8,7 @@ from tempfile import TemporaryDirectory
 import pytest
 
 runner = CliRunner()
-unnamespaced_config_file_path = 'acceptance/pipeline_conf.yaml'
-namespaced_config_file_path = 'acceptance/namespaced_pipeline_conf.yaml'
+config_file_path = 'acceptance/pipeline_conf.yaml'
 
 
 def provider_config_file_path(execution_mode):
@@ -25,7 +24,7 @@ def test_cli_v1():
     with TemporaryDirectory() as tmp_dir:
         output_file_path = os.path.join(tmp_dir, 'pipeline.yaml')
 
-        result = runner.invoke(compiler.compile, ['--pipeline_config', unnamespaced_config_file_path, '--provider_config', provider_config_file_path('v1'), '--output_file', output_file_path])
+        result = runner.invoke(compiler.compile, ['--pipeline_config', config_file_path, '--provider_config', provider_config_file_path('v1'), '--output_file', output_file_path])
 
         assert result.exit_code == 0
         assert os.stat(output_file_path).st_size != 0
@@ -35,11 +34,7 @@ def test_cli_v1():
         assert workflow['kind'] == 'Workflow'
 
 
-@pytest.mark.parametrize("config_file_path, expected_pipelineinfo_name", [
-    (unnamespaced_config_file_path,  "test"),
-    (namespaced_config_file_path, "namespace-test")
-])
-def test_cli_v2(config_file_path: str, expected_pipelineinfo_name: str):
+def test_cli_v2():
     with TemporaryDirectory() as tmp_dir:
         output_file_path = os.path.join(tmp_dir, 'pipeline.yaml')
 
@@ -51,7 +46,7 @@ def test_cli_v2(config_file_path: str, expected_pipelineinfo_name: str):
         f = open(output_file_path, "r")
         pipeline = yaml.safe_load(f.read())
         assert pipeline['pipelineSpec']['schemaVersion'] == '2.0.0'
-        assert pipeline['pipelineSpec']['pipelineInfo']['name'] == expected_pipelineinfo_name
+        assert pipeline['pipelineSpec']['pipelineInfo']['name'] == "namespace-test"
 
 
 def test_failure():
--- a/argo/kfp-compiler/kfp_compiler/compiler.py
+++ b/argo/kfp-compiler/kfp_compiler/compiler.py
@@ -67,8 +67,10 @@ def load_fn(tfx_components: str, env: list):
 
     return fn
 
-def parseNamespacedPipelineName(namespacedName: str) -> str:
-    return namespacedName.replace("/", "-")
+
+def parse_namespaced_pipeline_name(namespaced_name: str) -> str:
+    return namespaced_name.replace("/", "-")
+
 
 @click.command()
 @click.option('--pipeline_config', help='Pipeline configuration in yaml format', required=True)
@@ -80,9 +82,11 @@ def compile(pipeline_config: str, provider_config: str, output_file: str):
         pipeline_config_contents = yaml.safe_load(pipeline_stream)
         provider_config_contents = yaml.safe_load(provider_stream)
 
-        click.secho(f'Compiling with pipeline: {pipeline_config_contents} and provider {provider_config_contents} ', fg='green')
+        click.secho(f'Compiling with pipeline: {pipeline_config_contents} and provider {provider_config_contents} ',
+                    fg='green')
 
-        pipeline_root, serving_model_directory, temp_location = pipeline_paths_for_config(pipeline_config_contents, provider_config_contents)
+        pipeline_root, serving_model_directory, temp_location = pipeline_paths_for_config(pipeline_config_contents,
+                                                                                          provider_config_contents)
 
         beam_args = provider_config_contents.get('defaultBeamArgs', [])
         beam_args.extend(pipeline_config_contents.get('beamArgs', []))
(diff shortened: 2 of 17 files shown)


argo/kfp-compiler/kfp_compiler/compiler.py
```diff
@@ -94,7 +100,7 @@ def compile(pipeline_config: str, provider_config: str, output_file: str):
 
         compile_fn(pipeline_config_contents, output_file).run(
             pipeline.Pipeline(
-                pipeline_name=pipeline_config_contents['name'],
+                pipeline_name=parse_namespaced_pipeline_name(pipeline_config_contents['name']),
```
grahamia, line 103: we need to do this conversion here as `/` isn't a valid character for pipeline name

TobyPinfold (dismissed):

TobyPinfold (approved):

argo/providers/vai/provider.go
```diff
@@ -193,8 +202,40 @@ func (vaip VAIProvider) DeletePipeline(ctx context.Context, providerConfig VAIPr
 	return nil
 }
 
+func ExtractFromMap[T any](targetMap map[string]any, fieldName string) (T, error) {
+	result, ok := targetMap[fieldName].(T)
+	if !ok {
+		err := errors.New("Failed extracting field " + fieldName + "from a given map")
```
peter-hazell, line 208: would there be any value including the map itself in the error message as well ?

argo/providers/vai/provider.go
```diff
@@ -193,8 +202,40 @@ func (vaip VAIProvider) DeletePipeline(ctx context.Context, providerConfig VAIPr
...
+		var emptyResult T
+		return emptyResult, err
+	}
+	return result, nil
+}
```
peter-hazell, line 213: think this can be simplified, as I understand it from reading `.(T)` is a type assertion and will assign the value to `result` if the type matches, otherwise `result` will be empty. So by naming the return values, you can just assign the error to `err` if `ok` is false (bcos result will be empty anyway)

saves you having to construct an empty value for `T`

```suggestion
func ExtractFromMap[T any](targetMap map[string]any, fieldName string) (result T, err error) {
	result, ok := targetMap[fieldName].(T)
	if !ok {
		err = fmt.Errorf("failed extracting field %s from the given map", fieldName)
	}
	return result, err
}
```

grahamia: Cool, yes, that works, have combined with the error message change

argo/kfp-compiler/kfp_compiler/compiler.py
```diff
@@ -68,6 +68,10 @@ def load_fn(tfx_components: str, env: list):
     return fn
 
 
+def parse_namespaced_pipeline_name(namespaced_name: str) -> str:
+    return namespaced_name.replace("/", "-")
```
aidandunlop, line 72: would this affect the name of the Pipeline in the eventSource? i.e. if we had a Pipeline resource called `my/pipeline`, would the associated Sensor that triggers when the pipeline runs successfully need to match on `my-pipeline` or `my/pipeline`?
If the latter then worth updating the docs that `/` should be avoided in pipeline names. No pipelines have done that yet but could see it happening with a variant i.e. `pipeline/variant-training`

TobyPinfold: I don't think its possible as K8s doesn't allow for `/` to be used, only `-` or `.`

aidandunlop: ok great

argo/kfp-compiler/kfp_compiler/compiler.py
```diff
@@ -132,8 +138,13 @@ def compile_v2(config: dict, output_filename: str):
...
 
 
+def name_from_namespaced_name(namespaced_name: str) -> str:
+    separator_index = namespaced_name.find("/")
+    return namespaced_name[separator_index + 1:]
+
+
 def pipeline_paths_for_config(pipeline_config: dict, provider_config: dict):
-    pipeline_root = provider_config['pipelineRootStorage'] + '/' + pipeline_config['name']
+    pipeline_root = provider_config['pipelineRootStorage'] + '/' + name_from_namespaced_name(pipeline_config['name'])
```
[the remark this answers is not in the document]
grahamia: keeping KFP provider function as is for this PR, will be a separate task to look at the "multi-user isolation" of KFP

argo/providers/stub/provider.go
```diff
@@ -71,31 +73,31 @@ func (s StubProvider) DeletePipeline(_ context.Context, providerConfig StubProvi
...
+	return verifyCreateCall(providerConfig, ResourceDefinition{resourceDefinition.Name, resourceDefinition.Version})
 }
 
 func (s StubProvider) DeleteRun(_ context.Context, providerConfig StubProviderConfig, id string) error {
 	return verifyDeleteCall(providerConfig, id)
 }
 
 func (s StubProvider) CreateRunSchedule(_ context.Context, providerConfig StubProviderConfig, resourceDefinition base.RunScheduleDefinition) (string, error) {
-	return verifyCreateCall(providerConfig, ResourceDefinition{resourceDefinition.Name, resourceDefinition.Version})
+	return verifyCreateCall(providerConfig, ResourceDefinition{common.NamespacedName{Name: resourceDefinition.Name, Namespace: providerConfig.Namespace}, resourceDefinition.Version})
```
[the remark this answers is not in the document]
grahamia: Initially it was because of the experiment and the stub provider and not having the namespace available in the resource but I think the issue is probably that we set the ExpectedInput incorrectly in the integration test for experiment to have a namespace. No work has been done to namespace experiments. Will need to look into it

argo/kfp-compiler/acceptance/test_compiler.py
```diff
@@ -8,7 +8,8 @@
 import pytest
 
 runner = CliRunner()
-config_file_path = 'acceptance/pipeline_conf.yaml'
+unnamespaced_config_file_path = 'acceptance/pipeline_conf.yaml'
```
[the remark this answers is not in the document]
grahamia: Not sure I understand this comment? We wanted to ensure that it worked with both possible variants

argo/kfp-compiler/kfp_compiler/compiler.py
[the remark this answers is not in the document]
grahamia: Agree. This has confused us on how our testing has worked. We did this specific function before rediscussing with yourself and Jonas the namespace being used within the bucket and not just the bucket for acls. Will look into and correct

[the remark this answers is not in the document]
TobyPinfold: This is fixed in [c419c4b](https://github.com/sky-uk/kfp-operator/pull/326/commits/c419c4b97c2037bade934feba2e9996d0f8efb30)

argo/providers/stub/provider.go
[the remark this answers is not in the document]
TobyPinfold: Fixed in [8f7d8ac](https://github.com/sky-uk/kfp-operator/pull/326/commits/8f7d8ac652bebd4e69b9f9f0f07ca0abf89b1c42)

argo/kfp-compiler/kfp_compiler/compiler.py
```diff
@@ -68,6 +68,10 @@ def load_fn(tfx_components: str, env: list):
     return fn
 
 
+def parse_namespaced_pipeline_name(namespaced_name: str) -> str:
```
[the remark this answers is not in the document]
TobyPinfold: Fixed in [c419c4b](https://github.com/sky-uk/kfp-operator/pull/326/commits/c419c4b97c2037bade934feba2e9996d0f8efb30)

argo/kfp-compiler/acceptance/test_compiler.py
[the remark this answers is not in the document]
grahamia: Have discussed with Jonas and agree that the non-namespaced version isn't required as once code is deployed it will never not be namespaced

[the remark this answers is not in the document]
TobyPinfold: removed test in [f352a02](https://github.com/sky-uk/kfp-operator/pull/326/commits/f352a02828fb792b85f2aec10af06350958933c2)

argo/providers/vai/provider.go
```diff
@@ -193,8 +202,38 @@ func (vaip VAIProvider) DeletePipeline(ctx context.Context, providerConfig VAIPr
 	return nil
 }
 
+func ExtractFromMap[T any](targetMap map[string]any, fieldName string) (result T, err error) {
```
jmendesky, line 205: How about calling this `ExtractTypedFromMap`? I wasn't quite sure what this was supposed to do when I first looked at this function.
We could also move this into a separate file as it is not specific to this provider but considering that we are only using it here it is fine for now

jmendesky: actually, what is the reason for making this public? imo it should either be public in a separate file or privaye here

argo/providers/vai/provider.go
```diff
@@ -225,9 +268,15 @@ func (vaip VAIProvider) CreateRun(ctx context.Context, providerConfig VAIProvide
 		return "", err
 	}
 
+	runId, err := extractPipelineNameFromSpec(pipelineJob.PipelineSpec.AsMap())
```
jmendesky, line 271: I may be wrong but this strikes me as a relatively costly operation as the entire struct needs to be traveresed and converted. Could we use `GetStruct()...GetString()...` here?

argo/providers/vai/provider.go
```diff
@@ -193,8 +202,38 @@ func (vaip VAIProvider) DeletePipeline(ctx context.Context, providerConfig VAIPr
...
+	}
+	runId, err := ExtractFromMap[string](pipelineInfo, "name")
+	if err != nil {
+		return "", err
+	}
+
+	return runId, nil
+}
+
+func extractRunNameSuffix(pipelineName string) (string, error) {
```
jmendesky, line 227: I can't see where this is used apart from the tests 🙈

argo/providers/vai/provider.go
grahamia: Mistake at moment is that it's public, will change, as you say if more usage will move it out

argo/providers/vai/provider.go
grahamia: At least it was tested 🐵

argo/providers/vai/provider.go
grahamia: Looking

grahamia: Changed to using the struct directly to extract the value.

argo/providers/vai/provider.go
```diff
@@ -193,8 +202,30 @@ func (vaip VAIProvider) DeletePipeline(ctx context.Context, providerConfig VAIPr
...
+	value, ok := pbStruct.Fields[fieldName]
+	if !ok {
+		err = fmt.Errorf("failed extracting field %s from the given struct", fieldName)
+	}
+	return value, err
+}
+
+func extractPipelineNameFromPipelineSpec(ctx context.Context, pipelineSpec *structpb.Struct) (string, error) {
+	logger := common.LoggerFromContext(ctx)
+	pipelineInfo, err := extractFromStruct(pipelineSpec, "pipelineInfo")
```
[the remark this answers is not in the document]
grahamia: I didnt originally have the extra function calling function but it distracted too much to what was actually being achieved so like the extra function, also means that we can specifically test the extraction we want (i.e. pipelineInfo/name). I think a fully generic extraction function maybe a little overkill unless required elsewhere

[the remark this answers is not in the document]
jmendesky: ok that makes sense to me

PR #333, opened 2024-07-05
Use `nobody` user in custom alpine docker images

### LINKED ISSUE #65: Use non-root user in custom Dockerfiles
We currently create and run python-based images with user `root` only. This should be changed inline with container best practices.

TobyPinfold:
Closes #65

Adds a non-root user to docker images which are currently using a root user. It uses a userid of `65534`  `nobody` for the alpine images.

## Tasks

- [x] Update kfp compiler to use non-root user
- [x] Update stubbed provider to use non-root user
- [x] Documentation updated

### DIFF
--- a/argo/providers/stub/Dockerfile
+++ b/argo/providers/stub/Dockerfile
@@ -1,9 +1,10 @@
 FROM alpine:3.20.1
 
 WORKDIR /
+
 COPY inject.sh inject.sh
 COPY compile.sh compile.sh
 COPY bin/provider .
-USER 65532:65532
 
+USER 65534:65534
 ENTRYPOINT ["/bin/sh", "/inject.sh"]
--- a/argo/kfp-compiler/Dockerfile
+++ b/argo/kfp-compiler/Dockerfile
@@ -29,4 +29,5 @@ ADD resources/compile.sh kfp-compiler/compile.sh
 
 ADD resources/entrypoint.sh entrypoint.sh
 
+USER 65534:65534
 ENTRYPOINT ["/entrypoint.sh"]


aidandunlop (approved):

peter-hazell (approved):

grahamia (approved):
Lets looks at updating the base image versions in a new issue?

docs-gen/includes/quickstart/Dockerfile
paoloambrosio-skyuk, line 1: Probably not worth doing it here since it's meant to be as simple as possible not to confuse the users. It's an example and in a real pipeline they wouldn't copy the dataset either 😉

paoloambrosio-skyuk (commented):
Some General comments
- The Alpine image already comes with `nobody` (`65534`) and `nogroup` (`65533`) groups and a `nobody` (`65534`) user. Shouldn't we be using those?
- Also I'm not sure if it was intentional to have different Alpine base image versions, but probably not, and we should upgrade them to be the same.

docs-gen/includes/quickstart/Dockerfile
TobyPinfold: I agree with the above, I had missed the existing users (interesting alpine uses the slightly different userids for similar purposes), I'll update this PR to reflect this.

In terms of image upgrades, we have a separate PR upcoming for this.

TobyPinfold: Image upgrade available in https://github.com/sky-uk/kfp-operator/pull/335

argo/providers/stub/Dockerfile
```diff
@@ -1,9 +1,10 @@
 FROM alpine:3.20.1
 
 WORKDIR /
+
 COPY inject.sh inject.sh
 COPY compile.sh compile.sh
 COPY bin/provider .
-USER 65532:65532
```
TobyPinfold, line 7: originally used an ID that didn't exist, it now uses the consistent `nobody` user

grahamia (approved):
`noboby` approves

aidandunlop (approved):

Issue #128, opened 2022-06-16
Deleting a non-existing resource should succeed


jmendesky:
Delete of run schedule should succeed if the schedule has somehow been directly removed from within Vertex AI.

Currently if schedule doesnt exist the delete workflow fails and the resource stays in a state of "deleting"

grahamia:
Delete of run schedule should succeed if the schedule has somehow been directly removed from within Vertex AI.

Currently if schedule doesnt exist the delete workflow fails and the resource stays in a state of "deleting"

grahamia:
Tested deleting a pipeline after emptying gcs manually and delete is fine. It seams to just be the schedule issue.

PR #336, opened 2024-07-08
Fix delete run schedule not found with provider VAI

grahamia:
connects to #128 

On deleting a run schedule if Vertex AI responds with NotFound status then ignore the error as nothing to delete.

Without this change the resource was stuck in a "Deleting" status and never recovered.

Have checked with deleting pipelines and they got successfully deleted so no change needed for that.

### DIFF
--- a/argo/providers/vai/provider.go
+++ b/argo/providers/vai/provider.go
@@ -418,14 +418,18 @@ func (vaip VAIProvider) DeleteRunSchedule(ctx context.Context, providerConfig VA
 		Name: scheduleId,
 	})
 	if err != nil {
-		if status.Code(err) == codes.NotFound {
-			return nil
-		} else {
-			return err
-		}
+		return handleScheduleNotFound(err)
 	}
 
-	return deleteSchedule.Wait(ctx)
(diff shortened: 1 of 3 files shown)


TobyPinfold (approved):

argo/providers/vai/provider.go
```diff
@@ -415,10 +418,18 @@ func (vaip VAIProvider) DeleteRunSchedule(ctx context.Context, providerConfig VA
...
-	return deleteSchedule.Wait(ctx)
+	return handleScheduleNotFound(deleteSchedule.Wait(ctx))
+}
+
+func handleScheduleNotFound(err error) error {
+	if status.Code(err) == codes.NotFound {
+		return nil
+	} else {
+		return err
+	}
```
peter-hazell, line 432: don't need the else

can just `return err` after the if

grahamia: Done in https://github.com/sky-uk/kfp-operator/pull/336/commits/49525b221496cd1a53026ba84803dd8d55eb56a9

peter-hazell (approved):

edinekj (approved):

PR #339, opened 2024-07-09
Kubeflow Provider - handle NotFound in pipeline resource deletion

grahamia:
Connects to https://github.com/sky-uk/kfp-operator/issues/128

On deleting a pipeline resource with kubeflow provider when the pipeline didnt actually exist within kubeflow the deletion workflow failed and resource stayed in "deleting" status forever.

This change fixes that. On deleting check error to see if not found if so then ignore.

Refactored existing code in delete run schedule which was doing same logic.

### DIFF
--- a/argo/providers/kfp/provider.go
+++ b/argo/providers/kfp/provider.go
@@ -2,6 +2,7 @@ package kfp
 
 import (
 	"context"
+	"errors"
 	"github.com/argoproj/argo-events/eventsources/sources/generic"
 	"github.com/go-openapi/runtime"
 	"github.com/kubeflow/pipelines/backend/api/go_client"
@@ -90,6 +91,13 @@ func (kfpp KfpProvider) DeletePipeline(ctx context.Context, providerConfig KfpPr
 		ID:      id,
 		Context: ctx,
 	}, nil)
+
+	if err != nil {
+		var errorResult *pipeline_service.DeletePipelineDefault
+		if errors.As(err, &errorResult) && errorResult.Payload.Code == kfpApiConstants.KfpResourceNotFoundCode {
+			err = nil
+		}
+	}
 
 	return err
 }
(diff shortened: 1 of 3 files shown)


TobyPinfold (approved):

argo/providers/kfp/provider.go
```diff
@@ -91,6 +92,13 @@ func (kfpp KfpProvider) DeletePipeline(ctx context.Context, providerConfig KfpPr
 		Context: ctx,
 	}, nil)
 
+	if err != nil {
+		var errorResult *pipeline_service.DeletePipelineDefault
```
jmendesky, line 96: I like how this has been done for run schedules: https://github.com/sky-uk/kfp-operator/pull/336/files#diff-b10da3f39a3b0dc026835d1b8510861f3b8188618d9e35b550b80bd8cc4191cbR421

alexgeorgousis (approved):

argo/providers/kfp/provider.go
```diff
@@ -91,6 +92,13 @@ func (kfpp KfpProvider) DeletePipeline(ctx context.Context, providerConfig KfpPr
...
+		if errors.As(err, &errorResult) && errorResult.Payload.Code == kfpApiConstants.KfpResourceNotFoundCode {
+			err = nil
```
jmendesky, line 98: I wouldn't re-assign the error and instead `return nil` (same below)

grahamia: I personally didnt like the multiple return points and to make that possible with golang need to use the mutability possibility...

jmendesky: I don't necessarily like it either, but I prefer it over mutable errors. It is very [go-y](https://medium.com/swlh/return-early-pattern-3d18a41bba8) though 😓

argo/providers/kfp/provider.go
TobyPinfold: Theres no common interface between the two error types we try to handle so a generic function is not possible even with constraints.  In order to move the very small amount of common code into a common function will not be of much use past having an expressive name.

jmendesky (approved):
Not ideal having to pass the code into `ignoreNotFound` but we have no other choice.

PR #347, opened 2024-07-24
Reduce Quickstart Pipeline image size

### LINKED ISSUE #346: Reduce Quickstart Pipeline image size
### Context
The image size of the quickstart pipeline (`docs-gen/includes/quickstart/Dockerfile`) is getting quite large and subsequent updates to the TFX version increase the size:

1.12 -> 1.13 increased it to 9GB
1.13 -> 1.15 increases it to 11.4GB

It is possible to just install the `tfx` python package in a slim linux base image, which should reduce the size of the pipeline image and speed up build time.

aidandunlop:
Closes #346 
The quickstart pipeline uses the TFX image as it's base, which is ~9GB and includes several extra tools which aren't needed by the quickstart pipeline. This PR uses a slim python image as the base instead, and installs the `tfx` python package and other dependencies directly which will reduce the image size dramatically.

Image size when using base TFX image: 
- 20 GB when built locally 
- 9GB (virtual size from [here](https://console.cloud.google.com/artifacts/docker/nbcu-disco-mgmt-003/us/mlops/kfp-operator-quickstart/sha256:6419e4cf7447a7e1c35ced9dc6187aadc3c9365bd7f5edf0584b8bb7a992aefb))

Image size after using `tfx` python package on slim python: 
- 2.46GB (when built locally)
- 784**MB** (virtual size from [here](https://console.cloud.google.com/artifacts/docker/nbcu-disco-mgmt-003/us/mlops/kfp-operator-quickstart/sha256:31d092a44189c651ef6acb8ea83dc5d15f9e4bb6fbc1f7c7eedadfbc120f07e9))

## Tasks
- [ ] Validate the pipeline runs successfully 
- [x] Documentation updated


docs-gen/includes/quickstart/Dockerfile
```diff
@@ -1,10 +1,27 @@
-# TFX build
-FROM tensorflow/tfx:1.13.0
+FROM python:3.10.12 as builder
 
 RUN mkdir /data
 RUN wget https://raw.githubusercontent.com/tensorflow/tfx/master/tfx/examples/penguin/data/labelled/penguins_processed.csv -P /data
 
+RUN pip install poetry==1.7.1
```
jmendesky, line 6: Can build a wheel file on the host and install that into the container instead similar to what we did for [the compiler](https://github.com/sky-uk/kfp-operator/blob/master/argo/kfp-compiler/Dockerfile#L4)?

It would mean we can use the host's python cache instead of the container's and there shouldn't be a need to compile on the target platform explicitly. It also means we probably don't need a virtual env?

docs-gen/includes/quickstart/pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[tool.poetry]
+name = "quickstart"
+version = "0.1.0"
+description = ""
+authors = ["Hephaestus Team <DL-Discovery-Hephaestus@sky.uk>"]
```
jmendesky, line 5: We shouldn't expose this sky internal team, please

docs-gen/includes/quickstart/pyproject.toml
```diff
@@ -0,0 +1,12 @@
...
+
+[tool.poetry.dependencies]
+python = "3.10.12"
+tfx = "1.14.0"
+shapely = "<2"
+kfp = "1.8.22"
+setuptools = "~=70.0"
```
jmendesky, line 12:
```suggestion
setuptools = "~=70.0"

```

docs-gen/includes/quickstart/Dockerfile
adriangay: ok, will look into this today...

docs-gen/includes/quickstart/pyproject.toml
adriangay: done

docs-gen/includes/quickstart/pyproject.toml
adriangay: done

docs-gen/includes/quickstart/Dockerfile
adriangay: As discussed, installing via wheels isn't any more efficient:
- pip dependency resolution still takes a very long time for `tfx` (docker build on a local machine did not complete for over 3 hours).
- cache improvements wouldn't help on CI

We need to create a virtual env if we want to take advantage of multistage builds. [This article](https://medium.com/@albertazzir/blazing-fast-python-docker-builds-with-poetry-a78a66f5aed0) is a fairly good approach.

We will raise a new issue to make the kfp-compiler image consistent with this approach.

TobyPinfold (approved):

Ankitha-Krisshnan (approved):

docs-gen/includes/quickstart/Dockerfile
```diff
@@ -1,10 +1,27 @@
...
+
+COPY pyproject.toml .
+COPY poetry.lock .
+
+RUN poetry config virtualenvs.create true
+RUN poetry config virtualenvs.in-project true
+RUN poetry install
```
jmendesky, line 13: can we use `--without dev` as suggested in the article?

docs-gen/includes/quickstart/Dockerfile
jmendesky: ok makes sense 👍

docs-gen/includes/quickstart/pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[tool.poetry]
+name = "quickstart"
+version = "0.1.0"
```
jmendesky, line 3: It would be great if this could be set dynamically (like in the compiler) but this doesn't need to happen in this PR

docs-gen/includes/quickstart/Dockerfile
jmendesky: I suppose we don't do it here because we have no dev dependencies 😅

aidandunlop: added `--no-root` in e437d31dec58b158038a13043857f7b673f935cb

jmendesky (approved):

aidandunlop:
OK to test

PR #351, opened 2024-08-02
Ensure VAI Pipeline Job Id is unique

### LINKED ISSUE #349: Ensembled triggered pipelines fail on second run
**Describe the bug**
On an ensembled pipeline triggered by another pipeline run

eg.
```
  triggers:
    runConfigurations:
    - other-runconfig
```

Once triggered a one off run resource is created to trigger the rebuild of the model, this only works first time, second run fails with
```
    message: 'rpc error: code = AlreadyExists desc = Pipeline job with the specified
      ID already exists.'
```

**To Reproduce**
Steps to reproduce the behavior.
- create pipeline with run configuration
- create pipeline that ensembles this pipeline with trigger like above
- run first pipeline, this will correctly create one off run which will work
- run again but this time the one off run will fail with above error

**Expected behavior**
Triggered one off run should always work

**Version and configuration**
KFP-Operator version 0.6.0-6-g24a4729
Using VAI provider with an ensembled run configuration

**Logs**
From the generated run KFP-Operator creates off the back of having ensembled run configuration:
```
    message: 'rpc error: code = AlreadyExists desc = Pipeline job with the specified
      ID already exists.'
```

**Additional context**
Think it is to do with the way the job `PipelineJobId` is set in the VAI Provider createRun using the version from the run definition means it doesnt change between runs if the spec is the same which it is here. Think bug was introduced in the namespace separation work

grahamia:
Closes #349

```
Name of one off run is now <namespace>-<run definition name (this is the generated name for the run resource)>-<version>
```

e.g. namespace-kfp-operator-quickstart-tvrmk-1564fd

### DIFF
--- a/argo/providers/vai/provider.go
+++ b/argo/providers/vai/provider.go
@@ -205,30 +205,8 @@ func (vaip VAIProvider) DeletePipeline(ctx context.Context, providerConfig VAIPr
 	return nil
 }
 
-func extractFromStruct(pbStruct *structpb.Struct, fieldName string) (value *structpb.Value, err error) {
-	value, ok := pbStruct.Fields[fieldName]
-	if !ok {
-		err = fmt.Errorf("failed extracting field %s from the given struct", fieldName)
-	}
-	return value, err
-}
-
-func extractPipelineNameFromPipelineSpec(ctx context.Context, pipelineSpec *structpb.Struct) (string, error) {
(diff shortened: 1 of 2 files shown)


grahamia:
ok to test

jmendesky (approved):

aidandunlop (approved):

jmendesky (approved):

PR #352, opened 2024-08-06
Add validation method for RunCompletionEvent

edinekj:
# Summary:
Validator library added - https://pkg.go.dev/github.com/go-playground/validator/v10@v10.22.0

Added a validation extension method for `RunCompletionEvent` to validate the following fields:
- Status
- PipelineName
- RunId
- Either RunConfigurationName or RunName
- Provider 

Note: `RunConfigurationName` and `RunName` can have an empty Namespace field.

Note: Refactored the existing `suite_unit_test` to extract existing tests into own unit_test files.  So `common_types_unit_test.go` is a cut and paste from that file.

# Reason:
The Go JSON unmarshaller isn't strict, and any missing or extra fields are simply ignored. There is no way to add required fields using the Go JSON library.


argo/common/run_completion.go
```diff
@@ -20,18 +25,52 @@ var RunCompletionStatuses = struct {
...
 func (sre RunCompletionEvent) String() string {
 	return fmt.Sprintf("{Status:%s PipelineName:%+v RunConfigurationName:%+v RunName:%+v RunId:%s ServingModelArtifacts:%+v Artifacts:%+v Provider:%s}",
 		sre.Status, sre.PipelineName, sre.RunConfigurationName, sre.RunName, sre.RunId, sre.ServingModelArtifacts, sre.Artifacts, sre.Provider)
 }
+
+func validateNamespacedName(nn *NamespacedName, key string) error {
+	if nn == nil {
+		return fmt.Errorf("key: %s is nil", key)
+	} else if nn.Name == "" {
+		return fmt.Errorf("key: %s, Name filed is missing", key)
```
peter-hazell, line 48:
```suggestion
		return fmt.Errorf("key: %s, Name field is missing", key)
```

TobyPinfold (approved):

argo/common/run_completion.go
peter-hazell, line 48: Looks like there is still a typo??

edinekj: No idea how it has been missed even if we commit suggestions.

edinekj: 328d1937212e144ecb6c10f47f5f81b38edfd5c3

argo/common/run_completion.go
```diff
@@ -20,18 +25,52 @@ var RunCompletionStatuses = struct {
...
-	Status       RunCompletionStatus `json:"status"`
-	PipelineName NamespacedName      `json:"pipelineName"`
+	Status       RunCompletionStatus `json:"status" validate:"required"`
+	PipelineName NamespacedName      `json:"pipelineName" validate:"pipelineName"`
 	// Optionally render structs until https://github.com/golang/go/issues/11939 is addressed
-	RunConfigurationName  *NamespacedName `json:"runConfigurationName,omitempty"`
-	RunName               *NamespacedName `json:"runName,omitempty"`
-	RunId                 string          `json:"runId"`
+	RunConfigurationName  *NamespacedName `json:"runConfigurationName"`
+	RunName               *NamespacedName `json:"runName"`
```
grahamia, line 32: These should still be omitempty.

The validation here should be exactly one should be populated (with at least a name (validate both can be a separate card in future)).

if both populated - error
if neither populated - error
if one populated then all good.

argo/common/run_completion.go
```diff
@@ -20,18 +25,52 @@ var RunCompletionStatuses = struct {
...
+
+	return nil
+}
+
+func (sre RunCompletionEvent) Validate() error {
+	validate := validator.New()
+	validate.RegisterValidation("pipelineName", pipelineNameValidator)
+	validateErr := validate.Struct(sre)
+
+	runConfigurationNameValidationErr := validateNamespacedName(sre.RunConfigurationName, "RunCompletionEvent.RunConfigurationName")
```
grahamia, line 59: this needs changing following the previous comment

argo/common/suite_unit_test.go
```diff
@@ -5,9 +5,10 @@ package common
...
 	"encoding/json"
 	"fmt"
+	"testing"
+
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"testing"
 )
 
 func TestCommonUnitSuite(t *testing.T) {
```
grahamia, line 14: This test suite doesn't follow the same pattern as all the others, i.e. having just the suite setup in this suite file and then the tests for each component in own unit_test files. So split the existing namespacename tests out into own file and create a run_completion_event_unit_test file

grahamia (changes requested):

argo/common/run_completion.go
TobyPinfold: changed in [f7076ff](https://github.com/sky-uk/kfp-operator/pull/352/commits/f7076ffe6e057088e9725fc23b568083c3e76cab)

argo/common/run_completion.go
TobyPinfold: changed in [f7076ff](https://github.com/sky-uk/kfp-operator/pull/352/commits/f7076ffe6e057088e9725fc23b568083c3e76cab)

argo/common/suite_unit_test.go
TobyPinfold: done in [6996d38](https://github.com/sky-uk/kfp-operator/pull/352/commits/6996d3803367bab1833668a9763c2db01a8fcc73)

argo/common/run_completion.go
```diff
@@ -20,18 +24,57 @@ var RunCompletionStatuses = struct {
...
+		return fmt.Errorf("key: %s is nil", key)
+	} else if nn.Name == "" {
+		return fmt.Errorf("key: %s, Name field is missing", key)
+	}
+
+	return nil
+}
+
+func (sre RunCompletionEvent) Validate() error {
+	validate := validator.New()
```
grahamia, line 54: This shouldnt be done here. validator.New() is thread safe (well according to the comments in the code) so should only be created once and used not created on each message

grahamia (changes requested):

argo/common/run_completion_event_unit_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+import (
+	"fmt"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Context("RunCompletionEvent.Validate", func() {
+
+	validate, _ = InitialiseValidation()
```
grahamia, line 13: should check the error here else if it does happen to fail all you will get in the test output is probably a panic

argo/common/run_completion_event_unit_test.go
```diff
@@ -0,0 +1,189 @@
...
+
+	namespaceName := NamespacedName{
+		Name:      "Name",
+		Namespace: "Namespace",
+	}
+
+	baseEvent := RunCompletionEvent{
```
grahamia, line 20: It is weird to me that the baseEvent is an invalid one... usually that would be the valid one and then you make it invalid in those tests?

argo/common/run_completion_event_unit_test.go
```diff
@@ -0,0 +1,189 @@
...
+		RunConfigurationName: &namespaceName,
+		RunName:              &namespaceName,
+		RunId:                "RunId",
+		Provider:             "Provider",
+	}
+
+	It("is valid when RunConfigurationName is missing and RunName is provided", func() {
+		invalidEvent := baseEvent
+		invalidEvent.RunConfigurationName = nil
+		Expect(invalidEvent.Validate(validate)).To(Succeed())
```
grahamia, line 32: because reading this it says an invalid event is valid 🤯

TobyPinfold: yup needs to change

argo/common/run_completion_event_unit_test.go
TobyPinfold: its because they are both set, we could make them both empty instead, which we will do now

argo/common/run_completion.go
TobyPinfold: Changed in [5bc6a9f](https://github.com/sky-uk/kfp-operator/pull/352/commits/5bc6a9f67412272114e686d4ac9d403c0dbb9ccb) which uses Mutex to control the creation of the validator and registering of custom validation rules

argo/common/run_completion_event_unit_test.go
TobyPinfold: changed in [e78e58f](https://github.com/sky-uk/kfp-operator/pull/352/commits/e78e58f3ef3c6464dd3d9c2fcce26544d39f5512)

argo/common/run_completion_event_unit_test.go
TobyPinfold: changed in [e78e58f](https://github.com/sky-uk/kfp-operator/pull/352/commits/e78e58f3ef3c6464dd3d9c2fcce26544d39f5512)

argo/common/run_completion.go
```diff
@@ -19,19 +24,73 @@ var RunCompletionStatuses = struct {
...
+	if nn == nil {
+		return fmt.Errorf("key: %s is nil", key)
+	} else if nn.Name == "" {
+		return fmt.Errorf("key: %s, Name field is missing", key)
+	}
+
+	return nil
+}
+
+func (sre RunCompletionEvent) Validate(validate *validator.Validate) error {
```
jmendesky, line 72: shouldn't this be a struct-level validator instead?

argo/common/common_types_unit_test.go
```diff
@@ -0,0 +1,120 @@
...
+			serialised, err := json.Marshal(namespacedName)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(string(serialised)).To(Equal(`"` + namespace + "/" + name + `"`))
+			deserialised := NamespacedName{}
+			Expect(json.Unmarshal(serialised, &deserialised)).To(Succeed())
+			Expect(deserialised).To(Equal(namespacedName))
+		})
+	})
+
+	When("pointer is provider", func() {
```
jmendesky, line 28: Do you mean
```suggestion
	When("pointer is provided", func() {
```
?

argo/common/common_types_unit_test.go
```diff
@@ -0,0 +1,120 @@
...
+package common
+
+import (
+	"encoding/json"
+	
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Context("Marshal NamespacedName", Serial, func() {
```
jmendesky, line 12: These seem unrelated? If we added them here for completeness we should say so in the PR description

argo/common/run_completion_event_unit_test.go
```diff
@@ -0,0 +1,188 @@
...
+			invalidNamespacedName := namespacedName
+			invalidNamespacedName.Name = ""
+			invalidNamespacedName.Namespace = ""
+			Expect(validateNamespacedName(&invalidNamespacedName, "")).To(Not(Succeed()))
+		})
+
+	})
+})
+
+var _ = Context("RunCompletionEvent.String", func() {
```
jmendesky, line 121: This also seems unrelated to this PR

argo/common/common_types_unit_test.go
grahamia: Sorry, this was from my suggestion to refactor the existing tests that were all in suite_unit_test and to follow the same pattern as everywhere else and have own test files

argo/common/common_types_unit_test.go
grahamia: See https://github.com/sky-uk/kfp-operator/pull/352/files/e78e58f3ef3c6464dd3d9c2fcce26544d39f5512#diff-351efeff6045b8ad01ff61963051f1b236fe93ba9e21243be970e12f79bf7003L34 🐒

jmendesky (commented):
Can we please link to the newly introduced library? It makes it easier for reviewers (and future reference) to understand.

argo/common/common_types_unit_test.go
grahamia: I have added a note to the description. 👍🏼

argo/common/common_types_unit_test.go
jmendesky: Hahaha OK 😆 
still 🧐

argo/common/run_completion.go
```diff
@@ -19,19 +24,73 @@ var RunCompletionStatuses = struct {
...
 }
 
+var (
+	validate *validator.Validate
+	mutex    sync.Mutex
+)
+
+func InitialiseValidation() (*validator.Validate, error) {
+	if validate == nil {
+		mutex.Lock()
```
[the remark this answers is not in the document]
edinekj: We stoped using `mutex` and start using `init function` ab2e384ccef689f93479e1463161e1622c828cb4

TobyPinfold:
This logic will no longer be used within here, the onus for validation will be on clients reading the `RunCompletionEvents` to perform validation on events to their own requirements. So closing this change.

Issue #329, opened 2024-07-03
Create new custom resource for provider config and remove ConfigMap


grahamia:
# Overview

Currently provider configuration and build of resources is all within the KFP-Operator helm install, this makes changing/adding new providers cumbersome and it should be split into its own custom resource which then controller for that will manage the creation/update/deletion of associated resources.

Work has already started on this on this branch: https://github.com/sky-uk/kfp-operator/tree/provider-crd

# Acceptance Criteria

- [x] The existing provider ConfigMap should no longer exist
- [x] A custom resource for the newly defined CRD should replace the ConfigMap
- [x] When a Pipeline is created/updated, the workflow uses the custom resource and not the ConfigMap
- [x] When a RunConfiguration is created/updated, the workflow uses the custom resource and not the ConfigMap
- [x] When a RunSchedule is created/updated, the workflow uses the custom resource and not the ConfigMap
- [x] When a Run is created/updated, the workflow uses the custom resource and not the ConfigMap

# Tasks

- [x] Define CRD (CustomResourceDefinition)
- [x] Custom resources will be created as part of the Operator helm chart (replacing the creation of the config map). Leave values for the provider CRD in the helm values.
- [x] Load configuration from the custom resource rather than mounting the config map
- [x] Configuration is passed from operator into workflow as raw data mounted in volume
- [x] Update documentation

# Technical Details

See:
- https://github.com/sky-uk/kfp-operator/blob/master/helm/kfp-operator/templates/providers.yaml
- https://github.com/sky-uk/kfp-operator/blob/112bea2dac1c29e3cb6796bc6b8ef4fb65a79117/helm/kfp-operator/values.yaml#L53

**Example CR:**
```
apiVersion: pipelines.kubeflow.org/v1alpha5
kind: Provider
metadata:
  name: provider-name (e.g. kfp/vai)
spec:
  image: provider-image-url:0.0.1
  executionMode: v1
  serviceAccount: kfp-operator
  defaultBeamArgs:
  - name: arg1
    value: value1
  pipelineRootStorage: file:///directory
  parameters:
    pipelineBucket: bucket-path
    etc: etc
```

grahamia:
Need to rework provider into own chart first

PR #353, opened 2024-08-07
Define new Provider CRD and use within operator

Ankitha-Krisshnan:
Links to #329 

## Description 

This adds the CustomResourceDefinition (CRD) for the provider and passes this through the KFP Operator to the workflows. The behaviour from the workflows to the providers remains the same. This set of changes is a step towards replacing the existing ConfigMap with the provider CRD.


controllers/pipelines/state_handler_test.go
```diff
@@ -207,10 +207,14 @@ var _ = Describe("State handler", func() {
...
 		}
-		commands := stateHandler.stateTransition(context.Background(), provider, st.Experiment)
+		commands := stateHandler.stateTransition(context.Background(), *provider, st.Experiment)
 		Expect(commands).To(Equal(st.Commands))
-		if st.workflowFactory.CalledWithProvider != "" {
-			Expect(st.workflowFactory.CalledWithProvider).To(Equal(provider))
+		if st.workflowFactory.CalledWithProvider != nil {
+			Expect(st.workflowFactory.CalledWithProvider).To(BeComparableTo(provider))
+		} else {
+			fmt.Printf("BADGER SNAKE MUSHROOM %+v", provider)
```
Ankitha-Krisshnan, line 215: might want to change `BADGER SNAKE MUSHROOM`

apis/pipelines/v1alpha5/provider_types.go
```diff
@@ -0,0 +1,71 @@
...
+	ExecutionMode string `json:"executionMode" yaml:"executionMode"`
+	// +kubebuilder:validation:Pattern:=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+	ServiceAccount      string                           `json:"serviceAccount" yaml:"serviceAccount"`
+	DefaultBeamArgs     []apis.NamedValue                `json:"defaultBeamArgs,omitempty" yaml:"defaultBeamArgs,omitempty"`
+	PipelineRootStorage string                           `json:"pipelineRootStorage" yaml:"pipelineRootStorage"`
+	Parameters          map[string]*apiextensionsv1.JSON `json:"parameters,omitempty" yaml:"parameters,omitempty"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:resource:shortName="mlpr"
```
jmendesky, line 21: `mlpr` is not very intuitive. How about `mlprv` (like `srv` for `service`?)

apis/pipelines/v1alpha5/provider_types.go
```diff
@@ -0,0 +1,71 @@
...
+
+	Spec   ProviderSpec   `json:"spec,omitempty"`
+	Status ProviderStatus `json:"status,omitempty"`
+}
+
+type ProviderStatus struct {
+	Conditions Conditions `json:"conditions,omitempty"`
+}
+
+func (ps *ProviderStatus) SetSynchronizationState(state apis.SynchronizationState, message string, observedGeneration int64) {
```
jmendesky, line 38: This Code seems not to be used other than in the test? Do we need the `SynchronizationState` for this piece of work at all?

controllers/pipelines/pipeline_controller.go
```diff
@@ -56,7 +58,16 @@ func (r *PipelineReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 	desiredProvider := desiredProvider(pipeline, r.Config)
 
-	commands := r.StateHandler.StateTransition(ctx, desiredProvider, pipeline)
+	provider := pipelinesv1.Provider{
+		TypeMeta: metav1.TypeMeta{},
+		ObjectMeta: metav1.ObjectMeta{
+			Name: desiredProvider,
+		},
+		Spec:   pipelinesv1.ProviderSpec{},
+		Status: pipelinesv1.ProviderStatus{},
```
jmendesky, line 67: This creates an empty provider object and never reads it from the cluster. Have we missed that?

controllers/pipelines/resource_workflow_factory.go
```diff
@@ -105,12 +106,14 @@ func (workflows *ResourceWorkflowFactory[R, ResourceDefinition]) resourceDefinit
...
 }
 
-func (workflows *ResourceWorkflowFactory[R, ResourceDefinition]) ConstructCreationWorkflow(provider string, resource R) (*argo.Workflow, error) {
+func (workflows *ResourceWorkflowFactory[R, ResourceDefinition]) ConstructCreationWorkflow(provider pipelinesv1.Provider, resource R) (*argo.Workflow, error) {
 	resourceDefinition, err := workflows.resourceDefinitionYaml(resource)
 	if err != nil {
 		return nil, err
 	}
 
+	providerName := provider.ObjectMeta.Name
```
jmendesky, line 115: It looks like we are doing a roundtrip from provider name, to empty provider object, to name again...without ever reading it from the cluster?

controllers/pipelines/pipeline_controller.go
TobyPinfold: This is intended for this card, this is using the provider struct but for the moment only to carry the desiredProvider string name to the workflow. 

The next step is to actually load in the provider and populate these fields correctly

controllers/pipelines/resource_workflow_factory.go
TobyPinfold: same as above, this is a step towards loading the provider, we have separated the work into smaller PRs, the next step is to actually load the provider

apis/pipelines/v1alpha5/provider_types.go
TobyPinfold: removed in [a6ed438](https://github.com/sky-uk/kfp-operator/pull/353/commits/a6ed43815cfa649f714e9eb4e18d6785933c1513)

aidandunlop (approved):

apis/pipelines/v1alpha5/test_generators.go
```diff
@@ -32,6 +35,53 @@ func RandomPipelineSpec() PipelineSpec {
...
+		},
+		Spec:   RandomProviderSpec(),
+		Status: RandomProviderStatus(),
+	}
+}
+
+func RandomProviderSpec() ProviderSpec {
+	randomParameters := make(map[string]*apiextensionsv1.JSON)
+	for key, _ := range RandomMap() {
+		randomParameters[key] = &apiextensionsv1.JSON{Raw: []byte(`{"key1": "value1", "key2": 1234}`)}
```
peter-hazell, line 53: could randomise the keys and values?

TobyPinfold: this might overcomplicate the spec, its just a placeholder to check the params are working. Due to the JSON type it gets messy trying to template it

peter-hazell (approved):

PR #355, opened 2024-08-12
Provider CRD Update Workflows and Provider CLIs

TobyPinfold:
links to #329.

### Summary

This changes the current workflows by mounting the Provider CRD as an artifact and passing it through to the Provider CLI. Additionally, it updates the Provider CLI tools to accommodate the new structure of `provider-configs` to be used and includes changes to event source services to fetch providers by name from within a k8s cluster.

### Key Changes

1. **Workflow Enhancements:**
   - The provider config is passed through the workflow and mounted as an artifact for the Provider CLI tools to read.

2. **Provider CLI Updates:**
   - The Provider CLI tools have been updated to support the new structure of `provider-configs`.

3. **Event Source Serving Updates:**
   -  The Event Source serving for both KFP and VAI have been updated to read provider CRDs from a k8s cluster by name, rather than the existing approach of mounting a config map.

### Testing

- Verified all existing tests pass correctly for indistinguishable behaviour between using a config-map and provider CRDs
- Verified Creating and Updating and Deleting resources continues to work as expected.

### Documentation

- Updated relevant documentation to reflect changes from using config-maps to use provider CRDs and how to setup relevant RBAC to access provider CRDs


controllers/pipelines/suite_integration_test.go
```diff
@@ -26,9 +26,10 @@ import (
 )
 
 const (
-	TestTimeout   = 120
-	TestNamespace = "argo"
-	TestProvider  = "stub"
+	TestTimeout        = 120
+	TestNamespace      = "argo"
+	TestProvider       = "stub"
+	TestProviderConfig = `{"stub": "stub"}` // TODO: Update this to be real json
```
TobyPinfold, line 32:
```suggestion
```

grahamia (approved):

argo/providers/kfp/provider.go
```diff
@@ -398,24 +402,32 @@ func ConnectToKfpApi(address string) (*GrpcKfpApi, error) {
...
 	if err != nil {
 		return nil, err
 	}
 
-	metadataStore, err := ConnectToMetadataStore(providerConfig.GrpcMetadataStoreAddress)
+	config := &KfpProviderConfig{
+		Name: provider,
+	}
+
+	if err = LoadProvider[KfpProviderConfig](ctx, k8sClient, provider, namespace, config); err != nil {
```
aidandunlop, line 415: could the provider config be returned from `LoadProvider` explicitly? Bit confusing atm

grahamia: It's the go way, mutation. The config object needs to be created here as one is VAI and one is KFP specific, could just return as well as mutating in the function but then we would need to add the extra lines of if statement to handle the error response.

alexgeorgousis (approved):

aidandunlop (approved):

PR #371, opened 2024-10-03
Move artifact filtering to webhook

### LINKED ISSUE #366: Remove artifact filtering from provider event source servers and add to webhook (Step 2.5)
## Overview
currently the provider event source server filters job artifacts using information provided in the run configuration/run resource. This requires the provider to have to have specific KFP Operator knowledge and have access to KFP Operator resources which we don't want. Now we have the webhook within KFP Operator this filtering logic can be moved.

At this point kubeflow provider cant be done until  https://github.com/sky-uk/kfp-operator/issues/318 is fixed.

## Tasks
- [ ] remove the lookup of the run/runconfiguration from the event source server
- [ ] remove the filtering from the event source server
- [ ] pass all required information out (this will require a new struct to hold all information, extending RunCompletionEvent) from the event source server that KFP Operator will require to carry out the filtering, i.e.
    - [ ] task.TaskName
    - [ ] task.GetOutputs() - key from map, ie the name
    - [ ] list of all output artifacts, including uri and metadata (this will need to be serialised out as JSON and then deserialised back into the Struct on the KFP Operator side) for each
- [ ] logic needs to be added to load the run or run configuration within the webhook
- [ ] filter the passed artifacts the same as [provider ESS did](https://github.com/sky-uk/kfp-operator/blob/d3a38b0360fe198a2c2c831851919e6a0ad2e4f7/argo/providers/vai/eventing_server.go#L178)
- [ ] Output correct RunCompletionEvent struct

## Acceptance Criteria
- [ ] provider resources don't need read access to kfp-operator resources (i.e. run and runconfiguration)
- [ ] artifact list passed from provider into KFP Operator webhook should be unfiltered full list with all above information
- [ ] output produced onto the eventbus should now be the correct RunCompletionEvent structure with the filtered list of artifacts.
- [ ] any tests from ESS should be moved/rewritten in KFP Operator.

alexgeorgousis:
Closes #366 

## Tasks

- [ ] ...
- [ ] Documentation updated


argo/common/suite_unit_test.go
```diff
@@ -128,7 +128,7 @@ var _ = Context("NamespacedNameFromString", Serial, func() {
 var _ = Context("RunCompletionEvent.String", func() {
 	artList := []Artifact{
 		{
-			Name:     "ArtifactName",
+			Name:     "Name",
```
grahamia, line 131: need to revert this, think this has been done during a different refactor and intellij has done its thing

argo/common/test_utils.go
```diff
@@ -24,6 +27,32 @@ func RandomArtifact() Artifact {
 	return Artifact{Name: RandomString(), Location: RandomString()}
 }
 
+func RandomComponentArtifactInstance() ComponentArtifactInstance {
```
grahamia, line 30: dont think these should live here

controllers/webhook/event_processor.go
```diff
@@ -0,0 +1,110 @@
...
+			Namespace: runName.Namespace,
+			Name:      runName.Name,
+		}, runResource); err != nil {
+			logger.Error(err, "failed to load Run", "Run", runName)
+			return nil, err
+		}
+		return runResource.Spec.Artifacts, nil
+	} else {
+		logger.Error(errors.New("failed to retrieve resource artifacts"), "Neither RunConfigurationName or RunName specified")
+		return nil, errors.New("no RunConfiguration or RunName specified")
```
TobyPinfold, line 108: Any reason there are 2 different errors being created here, I feel like this could be 1 error used in both the return and the log.

controllers/webhook/event_processor_unit_test.go
```diff
@@ -0,0 +1,263 @@
...
+	groupVersion := schema.GroupVersion{Group: "pipelines.kubeflow.org", Version: "v1alpha5"}
+	scheme.AddKnownTypes(groupVersion, &v1alpha5.RunConfiguration{}, &v1alpha5.Run{})
+
+	metav1.AddToGroupVersion(scheme, groupVersion)
+	return scheme
+}
+
+func checkOutputArtifacts(outputArtifacts []v1alpha5.OutputArtifact, expectedArtifacts []v1alpha5.OutputArtifact) {
+	if outputArtifacts == nil {
+		Expect(expectedArtifacts).To(HaveLen(0))
```
TobyPinfold, line 31:
```suggestion
		Expect(expectedArtifacts).To(BeEmpty())
```

controllers/webhook/runcompletion_feed_unit_test.go
```diff
@@ -42,28 +43,82 @@ var _ = Context("getRequestBody", func() {
...
+
+			req, err := http.NewRequestWithContext(ctx, http.MethodPost, "http://example.com/events", bytes.NewReader([]byte("{\"hello\":\"world\"}")))
+			Expect(err).NotTo(HaveOccurred())
+			_, err = rcf.extractEventData(req)
+			Expect(err).To(MatchError("event data is empty"))
 		})
 	})
 
-	When("invalid body passed", func() {
+	//When("valid req")
```
TobyPinfold, line 105:
```suggestion
```

controllers/webhook/event_processor.go
grahamia: No reason apart from wording, have changed

grahamia (approved):
:shipit:

TobyPinfold (approved):

controllers/webhook/runcompletion_feed.go
```diff
@@ -63,16 +70,32 @@ func getRequestBody(ctx context.Context, request *http.Request) ([]byte, error)
...
 	if err != nil {
 		return nil, err
+	} else if rce == nil {
+		return nil, errors.New("event data is empty")
+	} else {
+		return &EventData{
+			Header:             request.Header,
+			RunCompletionEvent: *rce,
+		}, nil
 	}
```
alexgeorgousis, line 98: Minor but we typically split up if statements if they return. If you really prefer the if-elif-else then do that, up to you 👍 

```suggestion
	rce, err := rcf.eventProcessor.ToRunCompletionEvent(rcf.ctx, runDataWrapper.Data)
	if err != nil {
		return nil, err
	}
	
	if rce == nil {
		return nil, errors.New("event data is empty")
	}
	
	return &EventData{
		Header:             request.Header,
		RunCompletionEvent: *rce,
	}, nil
	}
```

grahamia: I think if ifelse else is better and less risk of introducing a bug in the execution flow

alexgeorgousis:
Latest changes look good ✅

PR #373, opened 2024-10-10
Update versions of `hugo` & `docsy` and enable dark mode selector

aidandunlop:
The hugo & docsy versions used by the website are quite out of date at the moment, and links such as "View page source" (on the right hand side of the page) are broken. The links contain `en/` in the path, but as we only support one language, this is an incorrect file path. Updating to the latest versions as below fixed these link errors.
- hugo upgraded from `extended_0.101.0` to `extended_0.135.0`
- docsy upgraded from `0.2.0` to `0.10.0`

This version of docsy also enables the ability to add a dark mode selector to the main right hand menu, meaning users can choose whether to view the site in `light`, `dark`, or `auto` mode.:

Screenshot 2024-10-10 at 13 07 14

There were various breaking changes when upgrading these versions:
- The syntax for reading code files with `readfile` to display code has changed, no longer needs to be wrapped inside a code block (if left inside the code block, the code is oddly rendered with extra whitespace)
- The `readfile` code also changed internally meaning it looked for files relative to the file that was importing the code, unless prefixed with `/`
- There was a missing closing `section` block, which now causes errors with the latest version of hugo
- Hugo only supports Goldmark as a markdown parser now, requiring a bit of config change

Other notable changes:
- Added nodejs 22.9.0 to tools-versions to ensure that node dependencies are installed (if the user doesn't have npm/node this wouldn't happen by default)
- config.toml renamed to hugo.toml in line [with the hugo conventions](https://gohugo.io/getting-started/configuration/#configuration-file):
  > With v0.109.0 and earlier the basename of the site configuration file was config instead of hugo. You can use either, but should transition to the new naming convention when practical.
- bump node size to `l-asdf`. It seems that the increase in dependencies means we run out of memory on the node prBuild/cdBuild executes on, and that causes prBuild to fail. Going to `m-asdf`

**Apologies for the large number of files changed, most of the website HTML/CSS/JS needed re-rendering / had updates from docsy**
## Tasks

- [x] hugo version updated
- [x] docsy version updated
- [x] Documentation updated


docs-gen/assets/scss/_styles_project.scss
```diff
@@ -0,0 +1 @@
+@import 'td/code-dark'
```
aidandunlop, line 1: this was required for dark mode. see [docsy docs](https://www.docsy.dev/docs/adding-content/lookandfeel/#lightdark-color-themes)

docs-gen/content/_index.md
```diff
@@ -42,3 +42,5 @@
 Uses the same great technologies like KFP itself.
 
 {{< /blocks/feature >}}
+
+{{< /blocks/section >}}
```
aidandunlop, line 46: Not having this closing section block now makes the build fail. It was just ignored previously

docs-gen/Makefile
```diff
@@ -15,3 +15,6 @@ clean:
 
 serve:
 	hugo server -b http://localhost:1313/
+
+install-npm-dependencies:
+	npm install
```
aidandunlop, line 20: This is to ensure postcss is installed (which is already in `package.json`). Without this we get the following error:
```
Error: error building site: POSTCSS: failed to transform "/scss/main.css" (text/css). You need to install PostCSS. See https://gohugo.io/functions/css/postcss/: binary with name "postcss" not found using npx
```

docs-gen/hugo.toml
```diff
@@ -39,5 +34,13 @@ canonifyURLs = true
...
 
+[params.ui]
+navbar_logo = false
+showLightDarkModeMenu = true
+
 [security.exec]
   osEnv = ['(?i)^((HTTPS?|NO)_PROXY|PATH(EXT)?|APPDATA|TE?MP|TERM|HOME|ASDF_DIR)$']
+
+[markup.goldmark.renderer]
+unsafe = true
```
TobyPinfold, line 45: For anyone wondering:

This doesn't inherently make the website unsafe. It just allows for the rendering of <html> elements from within markdown files.

docs-gen/assets/scss/_styles_project.scss
TobyPinfold, line 1:
```suggestion
@import 'td/code-dark'

```
pro-end of file new line to prevent diff pollution

TobyPinfold (approved):

docs-gen/hugo.toml
aidandunlop: yup, it's required now if you want to mix HTML and Markdown in the same file. It's described as 'unsafe' as it could allow others to execute JS from within the HTML if we were passing in user generated content, but as we control all of the content here, it's fine imo

aidandunlop: See https://gohugo.io/getting-started/configuration-markup/#rendererunsafe

aidandunlop:
This upgrade has caused some paths to JS and CSS files to contain an extra `/kfp-operator/`, e.g.:
```
href=https://sky-uk.github.io/kfp-operator/kfp-operator/scss/main.min.79e39.....etc.css
```
rather than:
```
href=https://sky-uk.github.io/kfp-operator/scss/main.min.79e39.....etc.css
```

I'm still not sure why, working on a fix but this shouldn't be merged yet

grahamia (approved):

docs-gen/layouts/partials/hooks/head-end.html
```diff
@@ -0,0 +1 @@
+<base href="{{ .Site.BaseURL }}">
```
aidandunlop, line 1: Fixing [the invalid resource paths](https://github.com/sky-uk/kfp-operator/pull/373#issuecomment-2407590392) by making all paths relative (setting `canonifyURLs` to false), meant that images could no longer be found - they needed to be referenced under `/kfp-operator/images/...etc`. Adding the baseURL in the head tag like we're doing in this file was [one recommended way of fixing this](https://discourse.gohugo.io/t/best-way-to-reference-an-image-in-static-directory-when-baseurl-could-contain-a-sub-directory/15461/3), and allows us to still reference images in markdown like so:
```markdown
![cd-ct](images/cd-ct.svg)
```

docs-gen/content/docs/examples/pipeline_dependencies.md
```diff
@@ -12,7 +12,7 @@ In this example, we break up the penguin example pipeline into two pipelines:
 - The *Penguin Examples Pipeline* has a single pipeline step that imports the CSV example and outputs it as an artifact
 - The *Penguin Training Pipeline* references the previously produces example and trains the model
 
-![ensembling.svg](/images/ensembling.svg)
+![ensembling.svg]({{< param "subpath" >}}/images/ensembling.svg)
```
aidandunlop, line 15: we have to do this for images now, as we're no longer setting `canonifyURLs` to true, so all links are relative. Without using the subpath param (defined in `hugo.toml`), this would try to get the image from `https://sky-uk.github.io/images/cd-ct.svg` rather than `https://sky-uk.github.io/kfp-operator/images/cd-ct.svg`

aidandunlop:
ok to test

TobyPinfold (approved):

docs-gen/Makefile
```diff
@@ -1,11 +1,5 @@
-build: clean
+build: clean install-npm-dependencies
 	hugo --minify
-	@# Workaround for https://github.com/google/docsy/issues/350 and https://github.com/gohugoio/hugo/issues/7714
-	$(eval URL_PATH = $(shell grep baseURL config.toml | sed "s/'//g" |  cut -d '/' -f 4))
-	@ if [ ! -z "$(URL_PATH)" ]; \
-	  then \
-	    sed -i.bak "s/background-image:url(/background-image:url(\/$(URL_PATH)/g" ../docs/index.html && rm ../docs/index.html.bak; \
-      fi
```
aidandunlop, line 8: we no longer need to do this as we're not setting `canonifyurls = true` to true anymore, meaning the background image already has the correct path

docs-gen/Makefile
```diff
@@ -14,4 +8,7 @@ clean:
 	find ../docs -mindepth 1 ! -name '.nojekyll' -delete
 
 serve:
-	hugo server -b http://localhost:1313/
+	hugo server -b http://localhost:1313/kfp-operator/
```
aidandunlop, line 11: this is preferable as it matches the sub path we have when deploying with github pages (and it's [what the README already says we should do](https://github.com/sky-uk/kfp-operator/blob/925290ccc5c63a158732d24b68f217ccd15edb84/docs-gen/README.md?plain=1#L13) anyway!)

bibcd.yml
```diff
@@ -3,8 +3,8 @@ context: mlops
 triggering: master-and-prs
 
 defaultNodes:
-  cdBuild: m-asdf
-  prBuild: m-asdf
+  cdBuild: l-asdf
+  prBuild: l-asdf
```
jmendesky, line 7: Why is this change needed?

docs-gen/hugo.toml
```diff
@@ -38,6 +33,15 @@ canonifyURLs = true
   github_subdir = "docs-gen"
   github_branch = "master"
   github_project_repo = "https://github.com/sky-uk/kfp-operator"
+  subpath = "/kfp-operator"
```
jmendesky, line 36: can we use this variable in `github_project_repo` and `baseURL`?

bibcd.yml
aidandunlop: prBuild often seems to get killed when installing python dependencies, often with the error message: 
```
Connection pool is full, discarding connection: pypi.org. Connection pool size: 10
```
Suspect it's running out of memory, hence the increase in node size

docs-gen/hugo.toml
aidandunlop: I can't think of a nice way to do it; TOML doesn't support variables directly

bibcd.yml
jmendesky: are we certain this solves the problem though? Also, this seems very unrelated to this PR and isn't even mentioned in the PR description.

aidandunlop: It appears to solve the problem, prBuild was flaky before I made that change and now it's not. I'm not sure how to prove it otherwise. I'll update the PR description though

jmendesky (approved):
Could you please get rid of obsolete commits by squashing them? we have a lot of intermediate commits (like setting the base tag) that are no longer needed for future reference.

PR #368, opened 2024-09-24
Operator run completion feed service

grahamia:
Closes #357 

A large chunk of the changes (all changes in pipelines etc) have come from a rename of the KfpControllerConfigSpec and changing "Workflow" to Spec to match convention everywhere else. Made this change here to add the extra configuration within the spec section (in workflow wouldn't have made sense)

## Tasks

- [x] new http service exposed
- [x] add configuration including default setup (no upstreams)
- [x] iterate over upstreams proxying the request on
- [x] test
- [x] Documentation updated

Issue #357, opened 2024-08-30
Add Webhook to KFP-Operator (Step 2)


grahamia:
## Overview
Currently provider sensor triggers http call to the Argo Event Webhook (within KFP Operator namespace) to publish the run completion event to the eventbus, this call should be moved to within the KFP Operator itself. As an interim step KFP Operator webhook will make a http request out to the existing Webhook (so acting as a kind of proxy at this point).

## Technical Details
Add new webhook to KFP Operator which will accept same message contents as existing Argo events Webhook. This is of the format:
```
{"header":{"Accept-Encoding":["gzip"],"Content-Length":["638"],"User-Agent":["Go-http-client/1.1"]},"body":{"specversion":"1.0","id":"33353061656164342d666137662d343233352d613365652d393531613334646637366330","source":"vai","type":"org.kubeflow.pipelines.run-completion","datacontenttype":"application/json","data":{"pipelineName":"mlops/kfp-operator-quickstart","provider":"vai","runConfigurationName":"mlops/kfp-operator-quickstart","runId":"mlops-kfp-operator-quickstart-mcsmf-31eab1-1-aidan-ed-webhook-test-14","runName":"mlops/kfp-operator-quickstart-mcsmf","servingModelArtifacts":[{"location":"gs://ml-kfp-operator-sandbox-100-pipelinestorage/mlops/kfp-operator-quickstart/serving/1726686874","name":"pushed_model"}],"status":"succeeded"}}}
```
At this stage since the webhook is just a proxy pass through body contents not to be checked and will just get passed onto upstream. Body will be checked and converted in [Step 2.5](#366)

The implementation should iterate through a list of endpoints to call. At this step this would be a http client calling the existing webhook. This will change in further step.

At this point the transmission of data to the webhook will be insecure and plaintext. Security and provider checks will be added in milestone 2.

Technically it wont be a webhook, it will be an api we expose through a new `api-service` similar to how the webhook service is exposed [here](https://github.com/sky-uk/kfp-operator/blob/81c9a040b071c0f2ec4cb98ef76f467adbb2e0b5/config/webhook/service.yaml). The http service will be started up as part of the main operator startup.

### Current Architecture
![image](https://github.com/user-attachments/assets/e883a896-4020-423d-8967-28de23acb5b5)

### After Architecture
<img width="1237" alt="image" src="https://github.com/user-attachments/assets/2abef03c-b3c1-4ba8-87d4-f7b07b795272">

## Acceptance Criteria
[...]

PR #377, opened 2024-10-17
Implement a gRPC event trigger to publish events to NATS

### LINKED ISSUE #358: Implement NATs event Trigger (Step 3)
## Overview
Currently the webhook call from KFP Operator is a http call to an Argo events webhook. 

This issue is to introduce a new component "NATs event trigger" with a gRPC interface, this service will process the request and produce the run completion message onto the eventbus.

## Technical Details
Deployment needs to be added to helm and kustomize.
gRPC interface will need to contain all the same information as is currently in the http payload
First task will be to defined the protobuf file for the service to include the following:
```
        "specversion": "1.0",
        "id": "123",
        "source": "vai",
        "type": "org.kubeflow.pipelines.run-completion",
        "datacontenttype": "application/json",
        "data": {
            "pipelineName": "pipelinename",
            "provider": "vai",
            "runConfigurationName": "runcomfigname",
            "runId": "runIdFromVAI",
            "runName": "runname",
            "servingModelArtifacts": [
                {
                    "location": "some_location",
                    "name": "pushed_model"
                }
            ],
            "status": "succeeded"
        }
```

Message put onto NATs eventbus should be same as current message.
Location of NATs eventbus should be configurable.

### Current Architecture
image

### After Architecture
image

## Acceptance Criteria
Eventing feedback should function as currently with no change
Testing should be carried out to ensure that same retry logic on failure exists and at least once consistency is maintained.

TobyPinfold:
Closes #358 

This adds a gRPC service which forwards run-completion events to a NATS event bus. This is currently planned for use with the webhook added in #357 to eliminate the need for argo sensors within the status feedback loop.

Changes:
- adds new gRPC service to pass run-completion events to a NATS event bus.
- adds the client integration to the existing controller webhook #368 to call this gRPC service.
- removes existing http client from controller webhook that calls an argo webhook sensor.
- removes argo webhook event source deployment for run-completion event feedback loop.


TobyPinfold:
ok to test

triggers/run-completion-event-trigger/dockerfile
chr12c, line 1: first letter uppercase for convention

triggers/run-completion-event-trigger/config/config.yaml
```diff
@@ -0,0 +1,8 @@
+natsConfig:
+  subject: events
+  serverConfig:
+    host: https://localhost
```
chr12c, line 4:
```suggestion
    host: localhost
```

triggers/run-completion-event-trigger/server/publisher_handler.go
```diff
@@ -0,0 +1,23 @@
...
+	Publish(data []byte) error
+}
+
+type NatsPublisher struct {
+	NatsConn *nats.Conn
+	Subject  string
+}
+
+func (nc NatsPublisher) Publish(data []byte) error {
+	log.Printf("Publish called")
```
chr12c, line 18:
```suggestion
	log.Print("Publish called")
```
nothing to format

triggers/run-completion-event-trigger/server/server.go
```diff
@@ -0,0 +1,36 @@
...
+type Server struct {
+	pb.UnimplementedRunCompletionEventTriggerServer
+	Config    *configLoader.Config
+	Publisher PublisherHandler
+}
+
+func (s *Server) ProcessEventFeed(_ context.Context, in *pb.RunCompletionEvent) (*emptypb.Empty, error) {
+	eventData, err := json.Marshal(in)
+	if err != nil {
+		err = status.Error(codes.InvalidArgument, "marshalling provided event failed")
```
chr12c, line 24: body description should have something to identify the rc

chr12c: same for the one below

triggers/run-completion-event-trigger/Makefile
```diff
@@ -0,0 +1,14 @@
...
+IMG := kfp-operator-run-completion-event-trigger
+
+build:
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -o bin/run-completion-event-trigger ./main.go
+
+test:
+	go test ./... -tags=unit
+
+functional-test: docker-build
+	docker compose up -d
```
chr12c, line 12:
```suggestion
functional-test:
        docker-build
	docker compose up -d
```

triggers/run-completion-event-trigger/README.md
```diff
@@ -0,0 +1,10 @@
+# Run Completion Event Trigger gRPC service
+
+## About:
+This adds a grpc service intended to pass run completion events to a nats service. This is generally used only by the
+events webhook built into the controller manager. 
+
+## Configuration:
+Configuration files `config.yaml` must be located in the same location the user is calling the binary from ie `$pwd` / `.`
+Alternatively a config.yml can be located at `/etc/run-completion-event-trigger/config.yaml` to be accessed when the binary is called
```
chr12c, line 9:
```suggestion
Alternatively a `config.yaml` can be located at `/etc/run-completion-event-trigger/config.yaml` to be accessed when the binary is called
```

triggers/run-completion-event-trigger/README.md
```diff
@@ -0,0 +1,10 @@
+# Run Completion Event Trigger gRPC service
+
+## About:
+This adds a grpc service intended to pass run completion events to a nats service. This is generally used only by the
+events webhook built into the controller manager. 
+
+## Configuration:
+Configuration files `config.yaml` must be located in the same location the user is calling the binary from ie `$pwd` / `.`
```
chr12c, line 8:
```suggestion
Configuration files `config.yaml` must be located in the same location the user is calling the binary from, i.e. `$pwd/.`.
```

triggers/run-completion-event-trigger/Makefile
TobyPinfold: I think as the docker-build is a direct call to another makefile definition as opposed to a CLI program it needs to be on the first line

controllers/webhook/upstream_services.go
```diff
@@ -1,114 +1,90 @@
...
-			req.Header.Add(headerName, headerValue)
-		}
+	runCompletionEvent := pb.RunCompletionEvent{
+		PipelineName:          pipelineName,
+		Provider:              event.Provider,
+		RunConfigurationName:  runConfigurationName,
+		RunId:                 event.RunId,
+		RunName:               runName,
+		ServingModelArtifacts: artifacts,
+		Status:                string(event.Status),
```
chr12c, line 86: why is rce treated differently from other strings which don't have what is effectively a string type alias

triggers/run-completion-event-trigger/server/suite_unit_test.go
```diff
@@ -0,0 +1,15 @@
...
+package run_completion_event_trigger
+
+import (
+	"testing"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+func TestRunCompletionEventTrigger(t *testing.T) {
```
chr12c, line 12: I think you can move this into `server_unit_test`

triggers/run-completion-event-trigger/proto/run_completion_event_trigger.proto
```diff
@@ -0,0 +1,26 @@
...
+}
+
+message RunCompletionEvent {
+  string pipeline_name = 1;
+  string provider = 2;
+  string run_configuration_name = 3;
+  string run_id = 4;
+  string run_name = 5;
+  repeated ServingModelArtifact serving_model_artifacts = 6;
+  string status = 7;
```
chr12c, line 20: we really going to use a string? Is there no exhaustive list of status coming back that we can check against

controllers/webhook/upstream_services_unit_test.go
```diff
@@ -0,0 +1,154 @@
...
+			ServingModelArtifacts: []common.Artifact{
+				{
+					Name:     "some-artifact",
+					Location: "gs://some/where",
+				},
+			},
+			Provider: "some-provider",
+		}
+
+		It("return no error when event data is converted to proto runcompletion event", func() {
```
chr12c, line 95:
```suggestion
		It("returns no error when event data is converted to proto runcompletion event", func() {
```

controllers/webhook/upstream_services.go
TobyPinfold: status is of `type RunCompletionStatus string` type alias  means we need to convert it back to its primitive `string` type to satisfy the compiler.   

`Artifact.Name` is already a `string` as below.

```
type Artifact struct {
	Name     string `json:"name"`
	Location string `json:"location"`
}
```

triggers/run-completion-event-trigger/server/suite_unit_test.go
TobyPinfold: should probably have a unit test for publishHandler for coverage even though its simple, will add

controllers/webhook/upstream_services.go
chr12c: ok. But given the other [comment](https://github.com/sky-uk/kfp-operator/pull/377#discussion_r1808797425), I guess this doesn't really matter. Can revisit if it does.

triggers/run-completion-event-trigger/config/config_unit_test.go
```diff
@@ -0,0 +1,65 @@
...
+				},
+			}
+			config, err := LoadConfig()
+			Expect(err).NotTo(HaveOccurred())
+			Expect(config).To(Equal(&expectedConfig))
+		})
+	})
+
+	When("given an enviornment variable", func() {
+		It("correctly override config value", func() {
```
chr12c, line 45:
```suggestion
	When("given an environment variable", func() {
		It("correctly overrides the config value", func() {
```

triggers/run-completion-event-trigger/proto/converters.go
```diff
@@ -0,0 +1,55 @@
+package run_completion_event_trigger
+
+import "github.com/sky-uk/kfp-operator/argo/common"
+
+func ProtoRunCompletionToCommon(protoRunCompletion *RunCompletionEvent) (common.RunCompletionEvent, error) {
```
chr12c, line 5: I think the term "domain" makes more sense than "comon"

chr12c: this function name doesn't really fit in with the `ArtifactsConverter` and `StatusConverter` (or the other way round if you prefer).

triggers/run-completion-event-trigger/server/publisher_handler.go
```diff
@@ -0,0 +1,35 @@
...
+type ConnectionError struct {
+	Error error
+}
+
+type NatsPublisher struct {
+	NatsConn *nats.Conn
+	Subject  string
+}
+
+func (nc NatsPublisher) Publish(runCompletionEvent common.RunCompletionEvent) (*MarshallingError, *ConnectionError) {
```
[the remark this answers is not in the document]
TobyPinfold: If we return just an Error, theres no way of determining the cause of the error from the callers side. Given we want to return different responses wether it was a connection error or a marshalling error this gets around that constraint.

triggers/run-completion-event-trigger/proto/converters.go
TobyPinfold: We haven't used `domain` across this repo at all, it wouldn't make sense to just start using it here.

re: ArtifactsConverter and StatusConverter, they were initially private, they are now private again. I think it makes sense for the public function to be more descriptive.

triggers/run-completion-event-trigger/server/publisher_handler.go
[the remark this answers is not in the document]
chr12c: You can define separate error types and then have them implement the Error() method. Then the caller can do a case switch on the error

config/manager/controller_manager_config.yaml
```diff
@@ -16,4 +16,7 @@ spec:
   workflowTemplatePrefix: kfp-operator-
   runCompletionFeed:
     port: 8082
-    endpoints: []
+    endpoints:
```
grahamia, line 19: I think this should be [] still for a default install

helm/kfp-operator/values.yaml
```diff
@@ -51,13 +51,22 @@ manager:
   webhookServicePort: 9443
   runcompletionWebhook:
     servicePort: 8082
-    endpoints: []
+    endpoints:
```
grahamia, line 54: As config comment think this should be [] for default install, no event feedback by default

helm/kfp-operator/values.yaml
```diff
@@ -51,13 +51,22 @@ manager:
...
+      - host: kfp-operator-run-completion-event-trigger.kfp-operator-system
+        path: ""
+        port: 50051
 
 statusFeedback:
   enabled: false
-  webhook:
-    endpoint: '/'
-    port: 12000
+  runcompletionEventTrigger:
```
grahamia, line 61: I see webhook was wrong before but this shouldn't be inside statusFeedback, should be part of a triggers section or something and default again should be it isn't deployed.

triggers/run-completion-event-trigger/server/server.go
```diff
@@ -0,0 +1,44 @@
...
+
+	if err = s.Publisher.Publish(commonRunCompletionEvent); err != nil {
+		switch e := err.(type) {
+		case *MarshallingError:
+			fmt.Println("NotFoundError:", e.Message)
+			return nil, status.Error(codes.InvalidArgument, "failed to marshal event")
+		case *ConnectionError:
+			return nil, status.Error(codes.Internal, "publisher request to upstream failed")
+		default:
+			return nil, status.Error(codes.Internal, "unknown error occurred")
```
chr12c, line 37: think it's more like "unexpected" errors right? Because this might be conflated with an actual `UNKNOWN` error (error from a different address space not known to this space).

triggers/run-completion-event-trigger/proto/run_completion_event_trigger.proto
```diff
@@ -0,0 +1,32 @@
+syntax = "proto3";
+
+import "google/protobuf/empty.proto";
+
+package run_completion_event_trigger;
+
+option go_package = "github.com/sky-uk/kfp-operator/triggers/run-completion-event-trigger/proto/run-completion-event-trigger";
```
alexgeorgousis, line 7: What does this do?

alexgeorgousis: https://protobuf.dev/reference/go/go-generated/#package

OK I guess it's just needed by the Go gRPC generation tool

TobyPinfold: https://protobuf.dev/reference/go/go-generated/#package

triggers/run-completion-event-trigger/README.md
```diff
@@ -0,0 +1,10 @@
+# Run Completion Event Trigger gRPC service
+
+## About:
+This adds a grpc service intended to pass run completion events to a nats service. This is generally used only by the
+events webhook built into the controller manager. 
+
+## Configuration:
+Configuration files `config.yaml` must be located in the same location the user is calling the binary from, i.e. `$pwd/.`.
+Alternatively a `config.yaml` can be located at `/etc/run-completion-event-trigger/config.yaml` to be accessed when the binary is called
+from anywhere in the system.
```
alexgeorgousis, line 10:
```suggestion
## About

This is a gRPC service that publishes [run completion events](https://sky-uk.github.io/kfp-operator/docs/reference/run-completion/) to a NATS service. This is currently used only by the
events webhook built into the controller manager.

## Configuration

The service can be configured by creating a file called `config.yaml`, which can be located either in the same directory the binary is executed from (i.e. `$pwd/.`) or `/etc/run-completion-event-trigger/config.yaml`, in which case the binary can be executed from anywhere in the system.

You can see the default configuration in [config/config.yaml](https://github.com/sky-uk/kfp-operator/blob/39409a00cf9ce3358f16a9ed636db6195cf26b78/triggers/run-completion-event-trigger/config/config.yaml).
```

triggers/run-completion-event-trigger/README.md
alexgeorgousis, line 1: Re: Directory structure

I think we should follow the [golang standard](https://github.com/golang-standards/project-layout?tab=readme-ov-file#go-directories) directory structure, so we'd get something like this:

```
Dockerfile
Makefile
README.md
docker-compose.yaml
cmd/
  main.go
  config.go
  config_unit_test.go
  config.yaml
internal/
  converters.go
  converters_unit_test.go
  publisher_handler.go
  server.go
  server_unit_test.go
  server_functional_test.go
proto/
  run_completion_event_trigger.proto
  run_completion_event_trigger.pb.go
  run_completion_event_trigger_grpc.pb.go
```

We could have sub-packages under `internal`, e.g. `converters` and `server` if you want, but the above is roughly how I'd picture the structure.

alexgeorgousis: We'll at some point need to follow this directory structure for the operator itself (according to the [kubebuilder docs](https://book.kubebuilder.io/migration/manually_migration_guide_gov3_to_gov4.html?highlight=internal#migration-from-project-config-version-gov3-to-gov4) for migrating from v3 (our current version) to v4).

Makefile
```diff
@@ -31,6 +31,11 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
 	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
 
+generate-grpc:
```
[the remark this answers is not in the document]
TobyPinfold: somewhat disagree, in that it is required to generate the client code for the http webhook within the controller also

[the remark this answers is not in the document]
aidandunlop: agreed as might also use this to generate more grpc code in the future

aidandunlop (approved):

controllers/webhook/upstream_services.go
```diff
@@ -1,114 +1,107 @@
...
 
 type UpstreamService interface {
-	call(ctx context.Context, ed EventData) error
+	call(ctx context.Context, ed common.RunCompletionEvent) error
 }
 
-type ArgoEventWebhook struct {
-	Upstream config.Endpoint
-	Client   *http.Client
+type GrpcNatsTrigger struct {
```
grahamia, line 19: In the controller side we dont know anything about what the underlying trigger is going to do so no need to say anything about NATS here now, it is just a gRPC trigger

helm/kfp-operator/templates/eventing/status-updater.yaml
```diff
@@ -84,6 +84,6 @@ spec:
         parameters:
         - src:
             dependencyName: events
-            dataKey: body.data
+            dataKey: body
```
grahamia, line 87: hmmm think this might cause clients issues, might need to discuss this.

TobyPinfold: good point, I've changed the code to retain the data key when published to nats so existing functionality should match what is currently used

controllers/webhook/upstream_services.go
TobyPinfold: changed to be generic `GrpcTrigger`

helm/kfp-operator/templates/eventing/triggers/run-completion-event-trigger/config.yaml
```diff
@@ -0,0 +1,17 @@
+{{- if .Values.runcompletionEventTrigger.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "kfp-operator.fullname" . }}-run-completion-event-trigger
+  namespace: kfp-operator-system
```
grahamia, line 6: this shouldnt be hardcoded

triggers/run-completion-event-trigger/Makefile
```diff
@@ -0,0 +1,14 @@
...
+
+IMG := kfp-operator-run-completion-event-trigger
+
+build:
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) go build -o bin/run-completion-event-trigger ./cmd/main.go
+
+test:
+	go test ./... -tags=unit
+
+functional-test: docker-build
```
grahamia, line 11: is this called as part of the build process?

grahamia: Ignore, see it is

helm/kfp-operator/templates/eventing/triggers/run-completion-event-trigger/config.yaml
TobyPinfold: made configurable as per controller

grahamia (approved):

chr12c (approved):

PR #379, opened 2024-10-18
Update controller-gen version to v0.12.0

TobyPinfold:
links to #358 

In order to utilise more recent gRPC libraries we need to upgrade the version of controller-gen used. It currently is using version 0.4.1 which is many years out of date. 

This PR upgrades the version used to 0.12.0 which supports v1.27 k8s and above [See here for compatibility](https://github.com/kubernetes-sigs/controller-tools?tab=readme-ov-file#compatibility).  This should be acceptable given it is already an unsupported version [See here for k8s supported versions](https://kubernetes.io/releases/)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,8 +4,6 @@ include newline.mk
 
 # Image URL to use all building/pushing image targets
 IMG ?= kfp-operator-controller
-# Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:ignoreUnexportedFields=false"
 
 all: build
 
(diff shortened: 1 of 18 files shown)


Makefile
```diff
@@ -5,7 +5,7 @@ include newline.mk
 # Image URL to use all building/pushing image targets
 IMG ?= kfp-operator-controller
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
```
grahamia, line 7: Assume this comment isnt valid anymore

Makefile
```diff
@@ -5,7 +5,7 @@ include newline.mk
...
-CRD_OPTIONS ?= "crd:preserveUnknownFields=false"
+CRD_OPTIONS ?= "crd:ignoreUnexportedFields=false"
```
grahamia, line 8: As discussing directly, this change doesnt seem to matchup

TobyPinfold: removed `ignoreUnexportedFields=false` as its the default behaviour.

In order to upgrade a CRD from v1beta1 to v1 you would need to use the `preserveUnknownFields=true` flag.
So reasons why we should not be using it:
- It was deprecated as of 1.22 which is EOL from years ago.
- All kfp custom resources are already using `apiVersion: apiextensions.k8s.io/v1` so theres no need for us to upgrade any CRDs here.

grahamia: Makes total sense to me 👍

grahamia (approved):

alexgeorgousis (approved):
LFTM (looks fantastic to me)

PR #383, opened 2024-10-25
Make pipeline compile create json rather than yaml

### LINKED ISSUE #382: KFP Operator Support for TFX 1.15
**Describe the bug**
On trying to use tfx >= 1.15 compile pipeline workflow is fine and passes but on creating a run or runconfiguration against the pipeline that workflow fails with 
```
message: invalid character '#' looking for beginning of value
```
This seems to be when the pipeline is loaded from pipeline storage bucket.

**To Reproduce**
Us TFX 1.15.1 and deploy pipeline and runconfiguration resource

**Expected behavior**
Should both deploy successfully

**Version and configuration**
If applicable, add configuration details to help explain your problem.

**Logs**
Workflow has (not much else):
```
message: invalid character '#' looking for beginning of value
```

grahamia:
Closes #382 .

To test I changed the workflow and ran a pipeline workflow with 1.14 tfx and that worked fine. Then I upgraded quickstart to 1.15.1 and ran that through and again that worked fine so think this fixes the issue. Pipeline stored in json format and runs against it all worked successfully.


aidandunlop (approved):

adriangay (commented):
We may need to support both JSON and YAML, ie. <TFX 1.15 and <=TFX 1.15. This PR proves we can support YAML. What proves we can support both? The fact you did manual tests may be acceptable?

chr12c (approved):

grahamia:
> We may need to support both JSON and YAML, ie. <TFX 1.15 and <=TFX 1.15. This PR proves we can support YAML. What proves we can support both? The fact you did manual tests may be acceptable?

This change forces the output of the compile to be json always, correcting the name of the file, the reference.yaml was json

PR #393, opened 2024-11-11
Move status updater webhook

chr12c:
Closes # .

## Tasks

- [ ] ...
- [ ] Documentation updated


argo/providers/vai/eventing_server.go
```diff
@@ -77,7 +77,7 @@ func (es *VaiEventingServer) StartEventSource(source *generic.EventSource, strea
 			return
 		}
 
-		es.Logger.V(1).Info("sending run completion data", "event", event)
+		es.Logger.Info("sending run completion data", "event", event)
```
alexgeorgousis, line 80: The `V(1)` sets the verbosity to a lower priority than a normal Info log, so it wasn't showing up in the logs. I'm not sure if it had a purpose, but removing it makes the log show up, so 🤷

alexgeorgousis (approved):

controllers/webhook/status_updater.go
```diff
@@ -0,0 +1,133 @@
...
+	if event.RunConfigurationName != nil {
+		if err := su.completeRunConfiguration(ctx, event); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (su *StatusUpdater) completeRun(ctx context.Context, event common.RunCompletionEvent) error {
+	if event.RunName.Namespace == "" {
```
grahamia, line 66: This seems a random extra check? Dont do the same check in the run configuration function?

controllers/webhook/status_updater.go
```diff
@@ -0,0 +1,133 @@
...
+			return nil
+		}
+		return err
+	}
+
+	if completionState := completionStateForRunCompletionStatus(event.Status); completionState != nil {
+		run.Status.CompletionState = *completionState
+		if err := su.K8sClient.Status().Update(ctx, &run); err != nil {
+			if errors.IsNotFound(err) {
+				return nil
```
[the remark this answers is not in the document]
TobyPinfold: Pretty sure this is how it was before, its slightly confusing as its not checking if the error exists, but wether the Error is of `NotFound` type.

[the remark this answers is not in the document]
TobyPinfold: ahh I see you were referring to another comment, might just be for completeness in which case as opposed to actually being reachable code, @chr12c  any comment?

controllers/webhook/runcompletion_event_handler.go
```diff
@@ -0,0 +1,10 @@
+package webhook
+
+import (
+	"context"
+	"github.com/sky-uk/kfp-operator/argo/common"
+)
+
+type RunCompletionEventHandler interface {
+	handle(ctx context.Context, ed common.RunCompletionEvent) error
```
aidandunlop, line 9: why `ed` and not just `event`?
```suggestion
	handle(ctx context.Context, event common.RunCompletionEvent) error
```

controllers/webhook/status_updater.go
TobyPinfold: This is done in the run configuration function also (though it is grouped with `status == succeeded`) but same concept applies here without a namespace being set its could be completely non-deterministic which resource should be updated. ie. `pipeline-1` in `default` namespace and `pipeline-1` in `other` namespace.

controllers/webhook/status_updater.go
```diff
@@ -0,0 +1,133 @@
...
+		if err := su.completeRunConfiguration(ctx, event); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (su *StatusUpdater) completeRun(ctx context.Context, event common.RunCompletionEvent) error {
+	if event.RunName.Namespace == "" {
+		return nil
```
grahamia, line 67: Also think it would be useful to log here to know we are just throwing away all the events

controllers/webhook/status_updater.go
grahamia: Yes, was blind and didnt see matching check in run config. Understand reasoning, lets add some logging to the situation

controllers/webhook/status_updater.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+	if event.RunConfigurationName != nil {
+		if err := su.completeRunConfiguration(ctx, event); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (su *StatusUpdater) completeRun(ctx context.Context, event common.RunCompletionEvent) error {
```
grahamia, line 65: We have a mixture of pointer receiver methods and not, this is bad practice

controllers/webhook/status_updater.go
chr12c: this entire file was copy-pasted, with the only difference being now it's got a `handle` function.

controllers/webhook/status_updater.go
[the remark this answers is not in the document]
chr12c: this is exactly how it worked before. See this comment: https://github.com/sky-uk/kfp-operator/pull/393/files#r1848149901

controllers/webhook/status_updater.go
```diff
@@ -0,0 +1,133 @@
...
+	if err := su.K8sClient.Get(
+		ctx,
+		types.NamespacedName{
+			Namespace: event.RunConfigurationName.Namespace,
+			Name:      event.RunConfigurationName.Name,
+		},
+		&rc,
+	); err != nil {
+		if errors.IsNotFound(err) {
+			return nil
```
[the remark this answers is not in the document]
chr12c: 6db3bda71992825720a76256562ce1770eefdd71

controllers/webhook/status_updater.go
chr12c: 6db3bda71992825720a76256562ce1770eefdd71

controllers/webhook/status_updater.go
chr12c: 8873223cdbe716463093424c66e8e762392db3e5

grahamia: I guess since I assume k8client doesnt hold much state it isnt much to create a copy. Intellij seems to always default to pointer receivers. For this case I think what you have done is correct

grahamia (approved):

peter-hazell:
does this need some integration tests?

controllers/webhook/status_updater.go
```diff
@@ -0,0 +1,183 @@
...
+func (su StatusUpdater) completeRunConfiguration(
+	ctx context.Context,
+	event common.RunCompletionEvent,
+) error {
+	logger := log.FromContext(ctx)
+
+	if event.Status != common.RunCompletionStatuses.Succeeded ||
+		event.RunConfigurationName.Namespace == "" {
+		logger.Info(
+			"RunCompletionEvent's RunConfigurationName namespace was the empty string. Skipping.",
```
TobyPinfold, line 132:
```suggestion
			"RunCompletionEvent's RunConfigurationName namespace in empty. Skipping.",
```

chr12c: I don't think so, because `RunConfigurationName` is a ptr, therefore "empty" could mean nil, whereas empty string means `RunConfigurationName` was not nil, but the `Namespace` was the empty string.

TobyPinfold: Its clear  the `namespace` was empty and not the `RunConfigurationName` so I don't think the pointer issue is relevant in this case

chr12c: https://github.com/sky-uk/kfp-operator/pull/393/commits/d15e740eba89062895ea9f4aabc7d27ed6515859

TobyPinfold (approved):

PR #394, opened 2024-11-11
Move provider annotation to spec and required field

### LINKED ISSUE #345: Move provider annotation to spec of resource
Currently it is an annotation, but the requested provider can be moved to be a mandatory field in the spec of all the KFP-Operator custom resources.
* Run
* RunConfiguration (via the associated `Run` in Spec.Run)
* Pipeline
* RunSchedule
* Experiment

Before this card:
```
apiVersion: pipelines.kubeflow.org/v1alpha5
kind: Pipeline
metadata:
  name: penguin-pipeline
  annotations:
    pipelines.kubeflow.org/provider: kfp
spec:
  image: kfp-quickstart:v1
  tfxComponents: base_pipeline.create_components
  env:
  - name: TRAINING_RUNS
    value: 100
```

After this card:
```
kind: Pipeline
metadata:
  name: penguin-pipeline
spec:
  provider: kfp
  image: kfp-quickstart:v1
  tfxComponents: base_pipeline.create_components
  env:
  - name: TRAINING_RUNS
    value: 100
```

Note that it will be in `Spec.Run.Provider` in RunConfigurations, rather than just in `Spec.Provider`, as the RunConfiguration references a Run:
```
spec:
  run:
    provider: kfp
    pipeline: some-pipeline
  triggers:
    onChange:
    - pipeline
    schedules:
    - 0 0 12 * * *
```

This then enforces that a provider must be specified for the resource.

Add this change to the v1alpha6 version

The default provider can't be removed from the KFP Operator configuration till v1alpha7 is created and we guarantee that provider is always populated (i.e. we drop alpha5 support)

grahamia:
Closes #345 

## Tasks

- [X] provider added to crd in config and helm
- [X] added conversion logic
- [X] updated code to use spec provider value rather than annotation
- [X] update tests
- [x] Documentation updated


aidandunlop (approved):

controllers/pipelines/runschedule_workflow_integration_test.go
```diff
@@ -11,13 +11,13 @@ import (
 var _ = Context("Resource Workflows", Serial, func() {
 	workflowFactory := RunScheduleWorkflowFactory(config.KfpControllerConfigSpec{
 		DefaultExperiment:      "Default",
-		DefaultProvider:        "stub",
+		DefaultProvider:        "not-used",
```
chr12c, line 14: can you put a doc to say this is not used on the struct?

main.go
```diff
@@ -84,7 +84,7 @@ func main() {
 			os.Exit(1)
 		}
 	}
-
+	pipelinesv1alpha5.DefaultProvider = ctrlConfig.Spec.DefaultProvider
```
chr12c, line 87: worth putting a doc here to explain why it has to be set like this

apis/pipelines/v1alpha5/conversion.go
```diff
@@ -1,9 +1,21 @@
...
 
 import (
+	"github.com/sky-uk/kfp-operator/apis"
 	hub "github.com/sky-uk/kfp-operator/apis/pipelines/v1alpha6"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+var (
+	DefaultProvider string
+)
```
TobyPinfold, line 11:
```suggestion
var DefaultProvider string
```

main.go
TobyPinfold: still no inline doc here, would agree it would be useful, maybe a `TODO:` remove for v1alpha7 release?

TobyPinfold (approved):
Some minor suggestions, otherwise looks good to me.

apis/pipelines/v1alpha5/conversion.go
grahamia: Applied in https://github.com/sky-uk/kfp-operator/pull/394/commits/97c408776a235828d74af923cad516b52bbb4c20

main.go
grahamia: Added todo in https://github.com/sky-uk/kfp-operator/pull/394/commits/97c408776a235828d74af923cad516b52bbb4c20

PR #397, opened 2024-11-19
Fix integration tests not using the new Provider struct in v1alpha6.

chr12c:
Fix integration tests not using the new Provider struct in v1alpha6.


grahamia (approved):
:shipit:

PR #403, opened 2024-11-27
Ensure run completion events with no artifacts return an empty slice rather than nil

### LINKED ISSUE #402: `servingModelArtifacts`/`artifacts` are `nil` when there are no artifacts, rather than an empty slice
**Describe the bug**
Previously the `servingModelArtifacts`/`artifacts` fields in the Run Completion Events were an empty slice (`[]`). They are now coming through as `nil`, meaning client code that performs some operation now has to explicitly handle the artifacts being nil and not a list.

**To Reproduce**
- Run a pipeline that does not produce any artifacts
- Create an Argo Sensor that reads the run completion events and logs the `artifacts` field
- The artifacts field will be `nil`, rather than `[]`

**Expected behavior**
- The `artifacts`/`servingModelArtifacts` fields should be `[]`, rather than `nil`

**Version and configuration**
- v0.6.0 
- latest code (`d2513a984b4a102cdc24e9cbaa6d6ce3a95f43aa`)

aidandunlop:
Closes #402.

Ensures `ServingModelArtifacts` and `Artifacts` return an empty slice rather than nil.

Modified the helper functions which convert to and from the Artifact protobuf message. Converting from the protobuf definition in the run completion event trigger was the necessary function to change, but decided to keep both functions consistent with each other.

## Tasks
- [ ] ...
- [ ] Documentation updated

### DIFF
--- a/argo/common/run_completion.go
+++ b/argo/common/run_completion.go
@@ -27,7 +27,7 @@ type RunCompletionEvent struct {
 	RunName               *NamespacedName `json:"runName,omitempty"`
 	RunId                 string          `json:"runId"`
 	ServingModelArtifacts []Artifact      `json:"servingModelArtifacts"`
-	Artifacts             []Artifact      `json:"artifacts,omitempty"`
+	Artifacts             []Artifact      `json:"artifacts"`
 	Provider              string          `json:"provider"`
 }
 
--- a/controllers/webhook/runcompletion_event_trigger.go
+++ b/controllers/webhook/runcompletion_event_trigger.go
@@ -86,7 +86,7 @@ func RunCompletionEventToProto(event common.RunCompletionEvent) (*pb.RunCompleti
 }
 
 func artifactToProto(commonArtifacts []common.Artifact) []*pb.Artifact {
-	var pbArtifacts = []*pb.Artifact{}
+	pbArtifacts := []*pb.Artifact{}
 	for _, commonArtifact := range commonArtifacts {
 		pbArtifacts = append(pbArtifacts, &pb.Artifact{
 			Location: commonArtifact.Location,
--- a/controllers/webhook/runcompletion_event_trigger.go
+++ b/controllers/webhook/runcompletion_event_trigger.go
@@ -86,7 +86,7 @@ func RunCompletionEventToProto(event common.RunCompletionEvent) (*pb.RunCompleti
 }
 
 func artifactToProto(commonArtifacts []common.Artifact) []*pb.Artifact {
-	var pbArtifacts = make([]*pb.Artifact, 0)
+	var pbArtifacts = []*pb.Artifact{}
 	for _, commonArtifact := range commonArtifacts {
 		pbArtifacts = append(pbArtifacts, &pb.Artifact{
 			Location: commonArtifact.Location,
--- a/controllers/webhook/runcompletion_event_trigger.go
+++ b/controllers/webhook/runcompletion_event_trigger.go
@@ -3,6 +3,7 @@ package webhook
 import (
 	"context"
 	"fmt"
+
 	"github.com/sky-uk/kfp-operator/argo/common"
 	"google.golang.org/grpc/credentials/insecure"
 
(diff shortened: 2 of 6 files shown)


controllers/webhook/runcompletion_event_trigger_unit_test.go
```diff
@@ -117,6 +118,22 @@ var _ = Context("EventDataToPbRunCompletion", func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 
+		It("returns an empty slice for no artifacts when event data is converted to proto runcompletion event", func() {
```
peter-hazell, line 121:
```suggestion
		It("returns an empty slices when there are no artifacts", func() {
```

don't need the `when event data is converted to proto runcompletion event` as that's already in the `When` for the test

aidandunlop: done in 776746f912d61af47d69f5eec9c65c0366752a5e

argo/common/run_completion.go
```diff
@@ -27,7 +27,7 @@ type RunCompletionEvent struct {
 	RunName               *NamespacedName `json:"runName,omitempty"`
 	RunId                 string          `json:"runId"`
 	ServingModelArtifacts []Artifact      `json:"servingModelArtifacts"`
-	Artifacts             []Artifact      `json:"artifacts,omitempty"`
+	Artifacts             []Artifact      `json:"artifacts"`
```
aidandunlop, line 30: `Artifacts` is now always populated as an empty list, so no need for the `omitempty`

controllers/webhook/runcompletion_event_trigger_unit_test.go
```diff
@@ -117,6 +118,22 @@ var _ = Context("EventDataToPbRunCompletion", func() {
...
 
+		It("returns empty slices when there are no artifacts", func() {
+			rceWithoutArtifacts := rce
+			rceWithoutArtifacts.Artifacts = []common.Artifact{}
+			rceWithoutArtifacts.ServingModelArtifacts = []common.Artifact{}
+
+			protoRce, err := RunCompletionEventToProto(rceWithoutArtifacts)
+			expectedArtifacts := []*pb.Artifact{}
+
+			Expect(protoRce.Artifacts).To(Not(BeNil()))
```
grahamia, line 129: Don't think this check is necessary.
```suggestion
```

controllers/webhook/runcompletion_event_trigger_unit_test.go
```diff
@@ -117,6 +118,22 @@ var _ = Context("EventDataToPbRunCompletion", func() {
...
+			Expect(protoRce.Artifacts).To(Equal(expectedArtifacts))
+			Expect(protoRce.ServingModelArtifacts).To(Not(BeNil()))
+			Expect(protoRce.ServingModelArtifacts).To(Equal(expectedArtifacts))
+
+			Expect(err).NotTo(HaveOccurred())
```
grahamia, line 134: Normally put this first straight after the call

controllers/webhook/runcompletion_event_trigger.go
```diff
@@ -85,7 +86,7 @@ func RunCompletionEventToProto(event common.RunCompletionEvent) (*pb.RunCompleti
 }
 
 func artifactToProto(commonArtifacts []common.Artifact) []*pb.Artifact {
-	var pbArtifacts []*pb.Artifact
+	var pbArtifacts = []*pb.Artifact{}
```
grahamia, line 89:
```suggestion
	pbArtifacts := []*pb.Artifact{}
```

controllers/webhook/runcompletion_event_trigger_unit_test.go
aidandunlop: done in ed748638ca3535a869efeb8e7ab2088598fb64ad

grahamia (approved):
:shipit:

controllers/webhook/runcompletion_event_trigger_unit_test.go
```diff
@@ -114,12 +117,27 @@ var _ = Context("EventDataToPbRunCompletion", func() {
...
+
+		It("returns empty slices when there are no artifacts", func() {
+			rceWithoutArtifacts := rce
+			rceWithoutArtifacts.Artifacts = []common.Artifact{}
+			rceWithoutArtifacts.ServingModelArtifacts = []common.Artifact{}
+
+			protoRce, err := RunCompletionEventToProto(rceWithoutArtifacts)
 			Expect(err).NotTo(HaveOccurred())
+
+			expectedArtifacts := []*pb.Artifact{}
```
peter-hazell, line 130:
```suggestion
			emptySliceOfArtifacts := []*pb.Artifact{}
```

aidandunlop: done in f6752969d1d8488b7eaa89eb86bcc3c4b3a37521

TobyPinfold (approved):

Issue #359, opened 2024-08-30
Event Processor (remove argo events from provider) (Step 4)


grahamia:
## Overview
To simplify the eventing process within the provider we can drop the event source server/eventsource/sensor complexity and have one simpler single deployment which carries out the same functionality. The beauty of this is that now there is no dependency on argo events within the provider.

## Technical Details
In this step the deployment of the new component can stay within the provider helm chart, in future card the deployment will be made part of the controller itself with no need for provider helm anymore.

The DataWrapper in the KFP Operator webhook can be removed at this point as no argo events wrapping needed.

### Current Architecture
![image](https://github.com/user-attachments/assets/9b070158-7c3d-40f6-9e9c-ffc711119f8b)

### After Architecture
<img width="1159" alt="image" src="https://github.com/user-attachments/assets/9c3818f2-d9d8-42fd-8343-d61951f4fbf0">

## Acceptance Criteria
Eventing feedback should function as currently with no change
Testing should be carried out to ensure that same retry logic on failure exists and at least once consistency is maintained.

PR #408, opened 2024-12-05
Implement controller to manage Provider service Deployment

### LINKED ISSUE #332: Provider Controller
## Overview

Now that we have simplified our run completion events setup (#359) we can implement a controller to manage Provider custom resources (introduced in #329). The beauty of being down to one component in the provider layer (replacing our Argo Events setup with a single Deployment) is that we have no need to manage multiple resources, so we have no need for helm, and therefore the controller can apply the resource directly.

## Technical Details
- Permissions for the Provider controller should be added to the existing `roles.yaml`, so the temporary `manager-providers-viewer-rolebinding` `ClusterRoleBinding` can be removed

### Current Architecture
image

### After Architecture
image

## Acceptance Criteria
- [x] When applying a new provider custom resource, the provider service Deployment resource is created
- [x] When updating an existing Provider resource, the provider service Deployment is updated
- [x] When deleting a Provider resource, the provider service Deployment is deleted
- [x] If a provider service Deployment is updated by something other than the Provider controller, the Deployment is re-synced to the desired state specified in the Provider resource
- [x] If a provider service Deployment is deleted by something other than the Provider controller, a new provider service Deployment is created (similarly to how `ReplicaSets` re-create `Pods` when deleted)
- [x] The provider service Deployment must have a unique name (similar to how Deployments create pods with unique names)
- [ ] The Provider controller tracks the state of provider resources. The state consists of:
    - [x] `observedGeneration`
    - [x] `SynchronizationState`
    - [x] `Synchronized` condition
    - [ ] Events (as outlined in #74 and implemented for our other CRDs)
- [ ] The KFP Operator docs have been updated as needed

### LINKED ISSUE #444: Implement controller to manage Provider service deployments
## Overview

We need to implement a controller to manage Provider resources (see parent issue for details). As part of this issue we'll create a basic controller which is responsible for deploying the [provider service](https://github.com/sky-uk/kfp-operator/issues/359) `Deployment` and managing its lifecycle.

## Acceptance Criteria

- [x] When applying a new provider custom resource, the provider service Deployment resource is created
- [x] When updating an existing Provider resource, the provider service Deployment is updated
- [x] When deleting a Provider resource, the provider service Deployment is deleted
- [x] If a provider service Deployment is updated by something other than the Provider controller, the Deployment is re-synced to the desired state specified in the Provider resource
- [x] If a provider service Deployment is deleted by something other than the Provider controller, a new provider service Deployment is created (similarly to how `ReplicaSets` re-create `Pods` when deleted)
- [x] The provider service Deployment must have a unique name (similar to how Deployments create pods with unique names)
- [x] The provider Helm chart has been made redundant and removed from the codebase

alexgeorgousis:
Closes #444 

## Overview

This PR creates a controller to manage `Provider` custom resources. It's responsible for creating and managing a `Deployment` that runs the Provider Service we implemented in #359.

## Useful Links

- [Argo events eventsource controller code](https://github.com/argoproj/argo-events/blob/master/pkg/reconciler/eventsource/resource.go)
- [Deployment status docs](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#deployment-status)


apis/pipelines/v1alpha5/provider_types.go
```diff
@@ -27,12 +27,8 @@ type Provider struct {
...
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   ProviderSpec   `json:"spec,omitempty"`
-	Status ProviderStatus `json:"status,omitempty"`
-}
-
-type ProviderStatus struct {
-	Conditions Conditions `json:"conditions,omitempty"`
+	Spec   ProviderSpec `json:"spec,omitempty"`
+	Status Status       `json:"status,omitempty"`
```
alexgeorgousis, line 31: We previously used a dedicated `ProviderStatus` struct for the provider CRD status, but we didn't see the point in it because:

- It only has a single field, `Conditions`, which is part of the default `Status` struct already
- We don't set the provider status anywhere (except for tests, just so they compile)

So we removed `ProviderStatus` and just used the default `Status` struct.

apis/config/v1alpha6/kfp_controller_config_types.go
```diff
@@ -3,43 +3,52 @@ package v1alpha6
...
+	cfg.ControllerManagerConfigurationSpec `json:"controller,omitempty"`
 }
 
-type Endpoint struct {
-	Host string `json:"host,omitempty"`
-	Port int    `json:"port,omitempty"`
-	Path string `json:"path,omitempty"`
+type KfpControllerConfigSpec struct {
+	DefaultProvider        string                `json:"defaultProvider,omitempty"`
+	DefaultProviderValues  DefaultProviderValues `json:"defaultProviderValues,omitempty"`
```
alexgeorgousis, line 23: The addition of `DefaultProviderValues` is the only real change in this file

apis/pipelines/object_hasher.go
```diff
@@ -38,6 +40,15 @@ func (oh ObjectHasher) WriteMapField(value map[string]string) {
 	oh.WriteFieldSeparator()
 }
 
+func (oh ObjectHasher) WriteObject(obj metav1.Object) error {
```
alexgeorgousis, line 43: This is used in `provider_controller.go` to compute the Deployment hash.

apis/pipelines/utils.go
```diff
@@ -186,3 +188,12 @@ func Values[K comparable, V any](kvs map[K]V) []V {
 
 	return vs
 }
+
+func MapConcat[K comparable, V any](m1, m2 map[K]V) map[K]V {
```
alexgeorgousis, line 192: This is used in the provider controller to concatenate label maps.

config/manager/controller_manager_config.yaml
```diff
@@ -17,3 +17,30 @@ spec:
   runCompletionFeed:
     port: 8082
     endpoints: []
+  defaultProviderValues:
```
alexgeorgousis, line 20: These are Provider config values that we didn't want to set on the Provider resources themselves (e.g. we don't want users to have to specify number of replicas and resource limits for their provider service deployments, that should be hidden).

config/rbac/deployment_manager_role.yaml
```diff
@@ -0,0 +1,18 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: deployment-manager-role
```
alexgeorgousis, line 5: This is used to give the provider controller permissions to manage the provider service deployment.

controllers/pipelines/provider_controller.go
alexgeorgousis, line 1: This is the main file to review and make sure it makes sense.

apis/config/v1alpha6/kfp_controller_config_types.go
```diff
@@ -3,43 +3,52 @@ package v1alpha6
...
+	WorkflowNamespace      string                `json:"workflowNamespace,omitempty"`
+	Multiversion           bool                  `json:"multiversion,omitempty"`
+	DefaultExperiment      string                `json:"defaultExperiment,omitempty"`
+	RunCompletionTTL       *metav1.Duration      `json:"runCompletionTTL,omitempty"`
+	RunCompletionFeed      ServiceConfiguration  `json:"runCompletionFeed,omitempty"`
 }
 
-func (e Endpoint) URL() string {
-	return fmt.Sprintf("%s:%d%s", e.Host, e.Port, e.Path)
+type DefaultProviderValues struct {
```
aidandunlop, line 32: might be pedantic, but should this be `DefaultProviderDeploymentValues`?

controllers/pipelines/provider_controller.go
```diff
@@ -0,0 +1,220 @@
...
+		},
+		Scheme: ec.Scheme,
+	}
+}
+
+//+kubebuilder:rbac:groups=pipelines.kubeflow.org,resources=providers,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=pipelines.kubeflow.org,resources=providers/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=pipelines.kubeflow.org,resources=providers/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+
```
aidandunlop, line 51: do we need this empty line? or does kubebuilder add it

controllers/pipelines/provider_controller.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	logger.Info("found provider", "resource", provider)
+
+	desiredDeployment, err := constructDeployment(provider, *r.Config.DeepCopy())
+	if err != nil {
+		logger.Error(err, "unable to construct provider service deployment")
+		return ctrl.Result{}, err
+	}
+	if err := ctrl.SetControllerReference(provider, desiredDeployment, r.Scheme); err != nil {
```
aidandunlop, line 70: should we log this err?

controllers/pipelines/provider_controller.go
```diff
@@ -0,0 +1,220 @@
...
+func (r *ProviderReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	provider := &pipelinesv1.Provider{}
+	return ctrl.NewControllerManagedBy(mgr).
+		For(provider).
+		Owns(&appsv1.Deployment{}).
+		Complete(r)
+}
+
+func constructDeployment(provider *pipelinesv1.Provider, config config.KfpControllerConfigSpec) (*appsv1.Deployment, error) {
+	matchLabels := map[string]string{AppLabel: fmt.Sprintf("provider-%s", provider.Name)}
```
aidandunlop, line 121: might be worth splitting out `fmt.Sprintf("provider-%s", provider.Name)` into a util func, is duplicated 3 times in this file. Could also use it in `GenerateName` inside Deployment ObjectMeta below, but you'll need to add the extra `-`

controllers/pipelines/provider_controller.go
```diff
@@ -0,0 +1,220 @@
...
+}
+
+func populateServiceContainer(serviceContainerName string, podTemplate v1.PodTemplateSpec, provider *pipelinesv1.Provider) v1.PodTemplateSpec {
+	for i, container := range podTemplate.Spec.Containers {
+		if container.Name == serviceContainerName {
+			podTemplate.Spec.Containers[i].Image = provider.Spec.ServiceImage
+			podTemplate.Spec.Containers[i].Env = append(podTemplate.Spec.Containers[i].Env, v1.EnvVar{
+				Name:  "PROVIDERNAME",
+				Value: provider.Name,
+			})
```
aidandunlop, line 165: could `break` here as we don't need to loop over the rest of the containers once you've found the service container

controllers/pipelines/provider_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	config "github.com/sky-uk/kfp-operator/apis/config/v1alpha6"
+	pipelinesv1 "github.com/sky-uk/kfp-operator/apis/pipelines/v1alpha6"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+var _ = Context("Provider Controller", func() {
```
aidandunlop, line 15: could add tests for `deploymentIsOutOfSync` and `syncDeployment`?

controllers/pipelines/provider_controller_test.go
```diff
@@ -0,0 +1,101 @@
...
+	var _ = Describe("constructDeployment", func() {
+		Specify("Should construct a Deployment with correct values given config and a Provider resource", func() {
```
aidandunlop, line 17: what about incorrect values? e.g. if `ServiceContainerName` did not match the correct container in `PodTemplateSpec`?

helm/kfp-operator/values.yaml
```diff
@@ -59,7 +59,15 @@ runcompletionEventTrigger:
...
+
+provider:
+  replicas: 1
+  resources:
+    requests:
+      cpu: 250m
+      memory: 128Mi
+    limits:
+      cpu: 500m
+      memory: 256Mi
```
aidandunlop, line 73: new line here?

controllers/pipelines/provider_controller.go
TobyPinfold, line 70:
```suggestion
	
	if err := ctrl.SetControllerReference(provider, desiredDeployment, r.Scheme); err != nil {
```

controllers/pipelines/provider_controller.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+
+	if existingDeployment != nil {
+		logger.Info("found existing provider deployment", "deployment", existingDeployment)
+
+		if err := setResourceHashAnnotation(existingDeployment); err != nil {
+			logger.Error(err, "unable to set resource hash annotation on existing deployment")
+			return ctrl.Result{}, err
+		}
+		if deploymentIsOutOfSync(existingDeployment, desiredDeployment) {
```
TobyPinfold, line 90:
```suggestion
		
		if deploymentIsOutOfSync(existingDeployment, desiredDeployment) {
```

controllers/pipelines/provider_controller.go
TobyPinfold: or just make it a var in here, 

```go
prefixedProviderName := fmt.Sprintf("provider-%s", provider.Name)
```

apis/config/v1alpha6/kfp_controller_config_types.go
alexgeorgousis: I think the idea is that those are default values for all Providers. Obviously, currently, Providers only use those values to pass them down to the Deployments they create, but:

1. That's an implementation detail
2. It might change in the future (there might be config values not directly used by the Deployment)

That's my thinking, let me know if it makes sense 🫡

controllers/pipelines/provider_controller.go
```diff
@@ -0,0 +1,220 @@
...
+	}
+	return podTemplate
+}
+
+func setResourceHashAnnotation(deployment *appsv1.Deployment) error {
+	hasher := pipelines.NewObjectHasher()
+	err := hasher.WriteObject(deployment)
+	if err != nil {
+		return err
+	}
```
TobyPinfold, line 177:
```suggestion
	if err := hasher.WriteObject(deployment); err != nil {
	    return err
	}
```

controllers/pipelines/provider_controller.go
alexgeorgousis: Thank you for your review and question about the blank line. I included it intentionally to enhance the readability, maintainability, and structure of the code. While seemingly minor, such formatting decisions can have meaningful impacts, especially in collaborative and long-term projects. Here are my reasons for keeping the blank line:

## 1. Readability
The blank line serves as a visual separator between logically distinct sections of the code. For example, if the line separates variable declarations, function definitions, or a block of configuration from the main logic, it makes the code easier to parse at a glance.
When someone revisits this code, especially a developer unfamiliar with its structure, the blank line helps guide their eyes and brain to group related blocks, reducing cognitive load.

## 2. Consistency with Style Guidelines
Many style guides, such as [PEP 8 for Python](https://peps.python.org/pep-0008/) or other language-specific conventions, often recommend blank lines to separate logical sections of code. This helps maintain a professional, clean format across the codebase.
Consistency in applying these practices ensures that all contributors write and read the code in a standardized manner.

## 3. Version Control and Changes
Blank lines make future diffs easier to understand. For instance, when adding or modifying code, the separation created by the blank line reduces the chance of unintended merging or overlapping of unrelated changes.
The line also helps clearly identify blocks of code that may need to be refactored or updated, improving the code's overall flexibility.

## 4. Error Prevention
Code written without deliberate separation can lead to mistakes. For instance, when functions or logic flow directly into one another with no visual break, it increases the likelihood of misunderstanding the boundaries of functionality, leading to bugs.

## 5. Best Practices in Team Environments
Teams often value clean, well-structured code because it promotes easier onboarding for new developers and faster comprehension for everyone involved. Including blank lines in strategic places adheres to this principle of writing code for humans first, compilers second.

## 6. Aesthetic and Organizational Value
While aesthetics may seem trivial, they play a significant role in making code approachable. A neatly organized file, with clear separations between sections, shows attention to detail and pride in the work.
This can be especially important in open-source or public-facing projects, where other developers frequently review and learn from the code.

In summary, the blank line might appear optional at first glance, but its inclusion provides meaningful benefits, especially for a team-oriented or long-term project. It improves readability, aligns with common standards, reduces the risk of errors, and enhances the code's maintainability. Unless there's a specific reason not to include it, I believe it adds value to the code.

That said, I’m open to discussing this further if the team prefers a different approach. 😊

controllers/pipelines/provider_controller.go
TobyPinfold: quite a few places within this file that are doing the same

controllers/pipelines/provider_controller.go
alexgeorgousis: Definitely, done in 8795e2e

controllers/pipelines/provider_controller.go
TobyPinfold: ahh see its used in some other funcs as well, agree that a new func makes sense

provider-service/base/pkg/config/config.go
```diff
@@ -57,3 +55,19 @@ func load() (*Config, error) {
...
+	viper.SetConfigType("json")
+	jsonBytes, err := json.Marshal(config)
+	if err != nil {
+		return err
+	}
+	reader := bytes.NewReader(jsonBytes)
+	err = viper.ReadConfig(reader)
+	if err != nil {
+		return err
+	}
```
TobyPinfold, line 71:
```suggestion
	if err = viper.ReadConfig(reader); err != nil {
		return err
	}
```

apis/config/v1alpha6/kfp_controller_config_types.go
aidandunlop: makes sense, especially changes in the future

controllers/pipelines/provider_controller.go
alexgeorgousis: I removed the `owner-name` label in 2204b32 because it was redundant. Now `fmt.Sprintf("provider-%s", provider.Name)` is only used once inside `constructDeployment`, plus the one in `GenerateName` that has the extra `-`.

So I'm thinking it's not worth extracting into a helper func anymore?

controllers/pipelines/provider_controller.go
alexgeorgousis: Good point, done in 631a298

controllers/pipelines/provider_controller_test.go
alexgeorgousis: - `deploymentIsOutOfSync`: Yeaaaah we could... but it's only 1 line of code, so the test would basically be re-writing the function and then calling it to test that it hasn't changed. Thoughts?
- `syncDeployment`: I added a unit test for this and it actually pointed out a bug - if the existing deployment has no annotations (because they were manually removed for example), then it panics because of nil pointer reference!

Done in [...]

controllers/pipelines/provider_controller.go
alexgeorgousis: Done in https://github.com/sky-uk/kfp-operator/commit/8b421e577d6992d542f07803bb55b3e9f1cea44a

controllers/pipelines/provider_controller.go
alexgeorgousis: Fixed all in https://github.com/sky-uk/kfp-operator/commit/8b421e577d6992d542f07803bb55b3e9f1cea44a

controllers/pipelines/provider_controller_test.go
alexgeorgousis: Hmmm not sure... should we throw an error in that case?

controllers/pipelines/provider_controller.go
alexgeorgousis: I created the `prefixedProviderName` var @TobyPinfold suggested in a609eb9

controllers/pipelines/provider_controller_test.go
alexgeorgousis: So... try to find a container with that name, if none is found return an error, otherwise mutate it to add the env vars etc.

aidandunlop: yeah I'd say we should throw an error as it would be a user misconfiguring the config and therefore they should be notified of such error at the earliest opportunity :)

controllers/pipelines/provider_controller.go
```diff
@@ -0,0 +1,219 @@
...
+		},
+		Scheme: ec.Scheme,
+	}
+}
+
+//+kubebuilder:rbac:groups=pipelines.kubeflow.org,resources=providers,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=pipelines.kubeflow.org,resources=providers/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=pipelines.kubeflow.org,resources=providers/finalizers,verbs=update
+//+kubebuilder:rbac:groups="",resources=events,verbs=create;patch
+
```
alexgeorgousis, line 49: It seems the blank line here isn't redundant after all (which is why we have it in all other controllers as well). `make manifests` doesn't pick up the provider-related resources it needs to generate if the blank line isn't there. It's a bit silly, but we need it 🤷‍♂️

controllers/pipelines/provider_controller_test.go
alexgeorgousis: Done in 7c1ad74

alexgeorgousis: FYI it would error anyway, because `constructDeployment` adds the `image` on the container, and if the `image` isn't set the Deployment is invalid.

But this is better because users will know why the image isn't set.

alexgeorgousis: FYI it would error anyway, because `constructDeployment` adds the `image` on the container, and if the `image` isn't set the Deployment is invalid.

But this is better because users will know why the image isn't set.

apis/pipelines/v1alpha6/provider_types.go
```diff
@@ -27,12 +17,39 @@ type Provider struct {
...
+// +kubebuilder:object:root=true
+type ProviderList struct {
+	metav1.TypeMeta `json:",inline"`
+	metav1.ListMeta `json:"metadata,omitempty"`
+	Items           []Provider `json:"items"`
+}
+
+type ProviderSpec struct {
+	ServiceImage  string `json:"serviceImage" yaml:"serviceImage"`
+	CliImage      string `json:"cliImage" yaml:"cliImage"`
```
grahamia, line 33: I think so this isnt a breaking change and to remove some changes from this pr (the workflows) and since cliImage wont actually be needed after the next provider api card is finished.

controllers/pipelines/provider_controller.go
```diff
@@ -0,0 +1,230 @@
...
+		if err := setResourceHashAnnotation(existingDeployment); err != nil {
+			logger.Error(err, "unable to set resource hash annotation on existing deployment")
+			return ctrl.Result{}, err
+		}
+
+		if deploymentIsOutOfSync(existingDeployment, desiredDeployment) {
+			logger.Info("resource hash mismatch, updating deployment")
+			existingDeployment = syncDeployment(existingDeployment, desiredDeployment)
+			if err = r.EC.Client.Update(ctx, existingDeployment); err != nil {
+				logger.Error(err, "unable to update provider service deployment", "deployment", desiredDeployment)
```
grahamia, line 94:
```suggestion
			desiredDeployment = syncDeployment(existingDeployment, desiredDeployment)
			if err = r.EC.Client.Update(ctx, desiredDeployment); err != nil {
				logger.Error(err, "unable to update provider service deployment", "deployment", desiredDeployment)
```
would it make more sense to replace value of `desiredDeployment` here?

apis/pipelines/v1alpha6/provider_types.go
alexgeorgousis: I re-read this a few times to understand what you're suggesting 😂

You mean rename `CliImage` back to `Image` so we don't have to change it anywhere, right? If yes, then yeah I like that. It's not the ideal name but like you said means: no breaking changes, no changes in the workflow templates, and it's gonna be removed soon anyway (before we release `v1alpha6`).

controllers/pipelines/provider_controller.go
alexgeorgousis: Well..... depends on how you conceptualise it I guess. The idea was we take the existing deployment in the cluster and update it based on the desired deployment. So the thing being "synced" is the existing deployment.

Kind of like how Argo CD Syncs an "existing" resource based on the "desired" version of that resource on your git repo.

apis/pipelines/v1alpha6/provider_types.go
alexgeorgousis: Done in 055227e

grahamia: Sorry, yes, my comment isn't great but you got it correct

controllers/pipelines/provider_controller.go
grahamia: I would say it reads better my way but am not precious.

TobyPinfold (approved):

aidandunlop (approved):

controllers/pipelines/provider_controller.go
alexgeorgousis: Alrighty I'll just keep it as is then ✅

grahamia (approved):

PR #409, opened 2024-12-08
Move controller LogKeys to internal package

### LINKED ISSUE #410: Move controller logging keys into an internal package
`logging.go` shouldn't be exposed outside of the `controllers/pipelines` package

Move the file into an internal package so that the compiler can hide this abstraction

chr12c:
Closes #410
Move controller LogKeys to internal package because it is not used outside of `controllers/pipelines`.


controllers/pipelines/internal/logging/logging.go
```diff
@@ -1,6 +1,6 @@
-package pipelines
+package logging
 
-var LogKeys = struct {
+var Keys = struct {
```
TobyPinfold, line 3: Could replace this struct with constants 
```go
const (
	Workflow  = "workflow"
	Command   = "command"
	Duration  = "duration"
	Status    = "status"
	OldStatus = "oldStatus"
	NewStatus = "newStatus"
)
```
Then just import the logging package with a name eg.

```go
import LogKeys "github.com/sky-uk/kfp-operator/controllers/pipelines/internal/logging"

LogKeys.Command
```

Benefits being:
- Smaller memory footprint
- immutable

chr12c: I did consider this way as well, but it means calling the package `logkeys` would make more sense, otherwise the callside would be `logging.Command`. `logkeys.Command` would show better intent. The other solution would be to keep the package as `logging`, but that would mean aliasing the import everysingle time to `logkeys`. Obviously just naming the package `logkeys` is preferable. Is that ok?

chr12c: or take it to the next level and have it as `logging.keys`. Then the import would be

```go
import "github.com/sky-uk/kfp-operator/controllers/pipelines/internal/logging"

foo := logging.keys.Command
```

TobyPinfold: I think `logkeys` would be best due not needing the named import

```go
import github.com/sky-uk/kfp-operator/controllers/pipelines/internal/logkeys
foo := logkeys.Command
```

chr12c: https://github.com/sky-uk/kfp-operator/pull/409/commits/9c6bb279e16f251836d9866db725e76fbce44d23

chr12c: https://github.com/sky-uk/kfp-operator/pull/409/commits/1902173a03c3d29ef00741ee4dd3846b33f6dd15

chr12c:
OK to test

TobyPinfold (approved):

chr12c:
OK to test

PR #411, opened 2024-12-10
Restructure controller pipelines package

### LINKED ISSUE #412: Restructure controller pipelines package to hide workflow factory
Put workflow factory related files into an internal package since these shouldn't be exposed outside of the pipelines package.

Put workflow factory related files into an internal package. Refactor as necessary.
Should be no functional changes.

chr12c:
Closes #412

## Tasks
- move WorkflowConstants and CommonWorkflowLabels from resource_workflow_factory to internal package (workflowconstants); prevents import cycle
- move workflow_utils to internal package (workflowutil)
- move test_shared to internal package (testutil)
- move workflow_test_helper to internal package (testutil)
- move all resource factories to internal package (workflowfactory)
- move RunConfigurationConstants from rc controller to run_workflow_factory; prevents import cycle


controllers/pipelines/internal/workflowutil/workflow_utils_test.go
```diff
@@ -1,14 +1,20 @@
...
 
 import (
 	argo "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"testing"
 )
 
+func TestWorkflowUtilsUnitSuite(t *testing.T) {
```
chr12c, line 13: Just an FYI in case anyone wants to do similar refactorings in the future: ginko hooks into a suite function, but since this test used to rely on the `suite_unit_test.go` on the `controllers/pipelines` package, it won't actually run this test and you won't know unless you look at what tests were ran. There are a few ways around this, e.g. register tests explicitly (see ginko docs). But I decided to make a new suite, which according to the docs is not bad practice either.

This is also why `workflowfactory` package has it's own unit and integration suites.

aidandunlop (approved):

grahamia (approved):

PR #419, opened 2024-12-19
Add `PublisherHandler` unit test

### LINKED ISSUE #418: Unit test for `PublisherHandler`
Add a unit test for the `PublisherHandler`

- The `PublisherHandler` unit test is tested.
- `NatsConn` will probably need to be mocked
- Don't use mockgen because it's archived

- testify
- mockery
- gomockery
- mockgen
**Additional context**

chr12c:
Closes #418 

Add `PublisherHandler` unit test and mock the nats connection using `testify`

## Tasks

- [x] Add unit test for `PublisherHandler`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ test-triggers:
 test-provider-service:
 	$(MAKE) -C provider-service test
 
-test-all: test helm-test-operator helm-test-provider test-argo test-triggers
+test-all: test helm-test-operator helm-test-provider test-argo test-triggers test-provider-service
 
 integration-test-all: integration-test
 	$(MAKE) -C argo/kfp-compiler integration-test
--- a/Makefile
+++ b/Makefile
@@ -73,6 +73,7 @@ integration-test-down:
 
 unit-test: manifests generate ## Run unit tests
 	go test ./... -tags=unit
+	$(MAKE) -C provider-service unit-test
 
 test: fmt vet unit-test decoupled-test
 
@@ -83,6 +84,9 @@ test-argo:
 
 test-triggers:
 	$(MAKE) -C triggers/run-completion-event-trigger test functional-test
+
+test-provider-service:
+	$(MAKE) -C provider-service test
 
 test-all: test helm-test-operator helm-test-provider test-argo test-triggers
 
--- a/triggers/run-completion-event-trigger/internal/publisher/publisher_handler.go
+++ b/triggers/run-completion-event-trigger/internal/publisher/publisher_handler.go
@@ -27,8 +27,12 @@ func (e *ConnectionError) Error() string {
 	return e.Message
 }
 
+type NatsConn interface {
+	Publish(subject string, data []byte) error
+}
+
 type NatsPublisher struct {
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -84,6 +85,9 @@ test-argo:
 test-triggers:
 	$(MAKE) -C triggers/run-completion-event-trigger test functional-test
 
+test-provider-service:
+	$(MAKE) -C provider-service test
+
 test-all: test helm-test-operator helm-test-provider test-argo test-triggers
```
TobyPinfold, line 91:
```suggestion
test-all: test helm-test-operator helm-test-provider test-argo test-triggers test-provider-service
```

aidandunlop (approved):

TobyPinfold (approved):

triggers/run-completion-event-trigger/internal/publisher/publisher_handler.go
```diff
@@ -27,16 +27,20 @@ func (e *ConnectionError) Error() string {
 	return e.Message
 }
 
+type NatsConn interface {
```
grahamia, line 30: This seems like a generic interface for publishing but is restricted by its name to NatsConn. Maybe just Publisher?

triggers/run-completion-event-trigger/internal/publisher/publisher_handler.go
grahamia, line 1: As this is just limited to Nats at the moment and not generic and New is explicitly returning a Nats implementation I think it should be called nats_publisher_handler.go or something

triggers/run-completion-event-trigger/internal/publisher/publisher_handler.go
chr12c: This interface is required in order to mock the behaviour of the nats conn. The real NatsConn isn't an interface so I'm relying on golang's duck typing to get enable the mocking.
`NatsPublisher.Publish` calls `NatsConn.Publish`.

triggers/run-completion-event-trigger/internal/publisher/publisher_handler.go
chr12c: renaming it back to `NewNatsPublisher`
https://github.com/sky-uk/kfp-operator/pull/419/commits/5235d4e57572834b8bdf681de6bc3dd247e7f2ea

triggers/run-completion-event-trigger/internal/publisher/publisher_handler.go
chr12c: https://github.com/sky-uk/kfp-operator/pull/419/commits/5235d4e57572834b8bdf681de6bc3dd247e7f2ea
I've made it lowercase for the moment so nothing outside can implement it

chr12c: I think if we do end up with multiple implementations of the `PublisherHandler`, they will split into different packages, with `natsConn` moving into the nats-only package.

grahamia (approved):

PR #425, opened 2025-01-02
Remove gomock from vai provider service eventflow, use testify

### LINKED ISSUE #424: Remove gomock from vai provider service eventflow
Remove gomock from eventflow because it's an archived library.
In https://github.com/sky-uk/kfp-operator/pull/422 it's starting to become unwieldly to have to setup so many things in mockgen for all of the new interfaces that have been introduced. But due to some existing tests already using gomock, it would be very untidy to run 2 mocking libraries in the same module.

Use testify instead. Is already used in the PublisherHandler

uber's fork of gomock. No need to look further because testify does everything we need from gomock without having to run mockgen.
- Not requiring mockgen setup in the Makefile
- but does tradeoff for having to write some simple boilerplate in test setup
- doesn't need a mock controller
- the remaining syntax is slightly less verbose
- is a maintained library that's quite popular

chr12c:
Closes #424

## Tasks

- [x] Remove gomock from vai provider service eventflow
- [x] Replace mocked client with testify


chr12c:
OK to test

provider-service/vai/internal/event_flow_unit_test.go
```diff
@@ -31,9 +32,21 @@ func artifact() *aiplatformpb.Artifact {
...
+	mock.Mock
+}
+
+func (m *MockPipelineJobClient) GetPipelineJob(
+	ctx context.Context,
+	req *aiplatformpb.GetPipelineJobRequest,
+	opts ...gax.CallOption,
+) (*aiplatformpb.PipelineJob, error) {
+	args := m.Called(ctx, req, opts)
+	return args.Get(0).(*aiplatformpb.PipelineJob), args.Error(1)
```
peter-hazell, line 45:
```suggestion
	return args.Get(0).(*aiplatformpb.PipelineJob), args.Error(1)
```

rather than always trying to cast to a `PipelineJob`, we should check if it's `nil` and return nil if so, this way we  don't have to pass an empty struct through just to satisfy the casting

chr12c: https://github.com/sky-uk/kfp-operator/pull/425/commits/0d8d38d8d638a3a5e5cc3ec8741b207d8112b763
https://github.com/sky-uk/kfp-operator/pull/425/commits/f338ac4486665ec039a129c69bc4162c3b1d0d8b

peter-hazell (approved):

aidandunlop (approved):

PR #429, opened 2025-01-10
Remove gomock from kfp provider. Replace with testify

### LINKED ISSUE #430: Remove gomock and mockgen from kfp provider service
As part of https://github.com/sky-uk/kfp-operator/pull/422, it would be convenient if gomock and mockgen was already removed since it's archived. That means we don't have to mock out the new interfaces with gomock just to change everything to testify afterwards, or change everything to testify in one go.
Although the latter is already done, that PR now has changes way out of scope, so the gomock part should be extracted into it's own PR to reduce clutter.

Remove gomock and mockgen from kfp provider service. Replace with testify

**Additional context**
already done in the vai provider service https://github.com/sky-uk/kfp-operator/issues/424

chr12c:
Closes #430

## Tasks

- [x] Remove gomock and mockgen from kfp provider service
- [x] Replace with testify
- [x] Move kfp provider config into it's own package


TobyPinfold (approved):
Looks good to me

peter-hazell (approved):

PR #431, opened 2025-01-10
Restructure vai provider service event flow

### LINKED ISSUE #432: Restructure vai provider service event flow
As part of https://github.com/sky-uk/kfp-operator/pull/422 and also golang project best practices, we should restructure the vai provider service's packages into something more browsable. Some of these ideas are in anticipation of the new server in that PR. Main goal is to declutter that PR.

Move eventing logic into separate package, as well as other bits like clients, mocks etc. which will be shared with the new server.

chr12c:
Closes #432

## Tasks
For vai provider-service:
- [x] Move event flow into separate pkg
- [x] Move labels into separate pkg
- [x] Move mocks and clients into separate pkgs
- [x] Update event flow test to check for pipeline job client's expected request


grahamia (approved):
Looks good after agreed change.

PR #433, opened 2025-01-10
Move provider service vai config's pipelineStorageObject and pipelineUri into util pkg

### LINKED ISSUE #434: Move provider service vai config's pipelineStorageObject and pipelineUri into util pkg
In the past, the vai provider config was copied over from the argo vai module to the provider-service event flow. However since the provider-service doesn't talk to gcs (until https://github.com/sky-uk/kfp-operator/pull/433), it didn't use all of the helper methods.

These methods will be used when the provider-service takes on the functionality of the argo CLI. Upon further inspection, a couple of these methods shouldn't be a method on the config since it doesn't use anything from the config. Instead they should be a standalone util.

Also when this config was copied over, the tests weren't copied over.

All of this was done in https://github.com/sky-uk/kfp-operator/pull/433, but should be in a separate PR for clarity.

put `pipelineStorageObject` and `pipelineUri` into a package so that it's separate from the vai provider config. Add tests and improve where necessary.

chr12c:
Closes #434 

## Tasks

- [x] Move provider service vai config's pipelineStorageObject and pipelineUri into util pkg
- [x] Copy (and update) test

### DIFF
--- a/provider-service/vai/internal/config/config.go
+++ b/provider-service/vai/internal/config/config.go
@@ -2,7 +2,6 @@ package config
 
 import (
 	"fmt"
-	"github.com/sky-uk/kfp-operator/argo/common"
 )
 
 type VAIProviderConfig struct {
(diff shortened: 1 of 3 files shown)


provider-service/vai/internal/util/util_test.go
```diff
@@ -0,0 +1,118 @@
...
+	. "github.com/onsi/gomega"
+	"github.com/sky-uk/kfp-operator/argo/common"
+)
+
+func TestGcsUtilUnitSuite(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "GCS Util Unit Suite")
+}
+
+var _ = Describe("Util", Ordered, func() {
```
chr12c, line 18: the equivalent test that exists in argo provider config: https://github.com/sky-uk/kfp-operator/blob/72df0a26515ec8c67167379610a3efa6459e6a1f/argo/providers/vai/internal/config_unit_test.go

grahamia (approved):

aidandunlop (approved):

provider-service/vai/internal/config/config.go
```diff
@@ -32,22 +31,6 @@ func (vaipc VAIProviderConfig) PipelineJobName(name string) string {
 	return fmt.Sprintf("%s/pipelineJobs/%s", vaipc.Parent(), name)
 }
 
-func (vaipc VAIProviderConfig) pipelineStorageObject(pipelineName common.NamespacedName, pipelineVersion string) (string, error) {
```
chr12c, line 35: yes, these are unused right now. Which is why the signature can be changed without breaking anything

PR #435, opened 2025-01-15
Shared provider service HTTP server

### LINKED ISSUE #436: Shared provider service HTTP server
Implement a new shared provider-service http server, where the sole responsibility of this server is to act as a router to then call a concrete provider implementation.

Has already been implemented in this branch: https://github.com/sky-uk/kfp-operator/pull/422
Since it can be isolated from the individual provider service implementations, it makes sense to separate the server part out into it's own PR.
For now, none of the providers will spin up this new http server.

chr12c:
Closes #436 
Implement shared provider service HTTP server and define server openapi spec

## Tasks

- [x] Implement shared provider service HTTP server
- [x] Documentation updated


provider-service/base/pkg/server/resource/experiment.go
```diff
@@ -0,0 +1,48 @@
...
+func (*Experiment) Type() string {
+	return "experiment"
+}
+
+func (e *Experiment) Create(body []byte) (ResponseBody, error) {
+	ed := ExperimentDefinition{}
+	err := json.Unmarshal(body, &ed)
+	if err != nil {
+		return ResponseBody{}, err
+	}
```
TobyPinfold, line 18:
```suggestion
	if err := json.Unmarshal(body, &ed); err != nil {
	    return ResponseBody{}, err
	}
```

chr12c: https://github.com/sky-uk/kfp-operator/pull/435/commits/49df6c316ebf617a6e68286025ab12c33f9b938e

TobyPinfold (approved):

provider-service/base/pkg/server/server.go
```diff
@@ -0,0 +1,134 @@
...
+			http.Error(w, "Failed to read request body", http.StatusInternalServerError)
+			return
+		}
+		defer r.Body.Close()
+
+		if err := hr.Update(decodedId, body); err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		w.WriteHeader(http.StatusOK)
```
grahamia, line 71: update also should be returning the id in output body as well as create

TobyPinfold: Good spot....does the id change on update though?

If it doesn't change it makes sense to just return 204 and not need the id be passed back as it would already be known by the client calling. 

If it does change then @chr12c we will need to update the api spec to return a 200 and response body

chr12c: It can change depending on what the underlying resource is. e.g. the kfp provider's UpdatePipeline will return the input id, and some of the other Updates will return a server generated id.

chr12c: 0f8fc3fb3913bf64f62a4a37fe927387ad458631

chr12c: https://github.com/sky-uk/kfp-operator/pull/435/commits/e6a8e5ebce9d8e610a079c6710b4f52932e6815b

TobyPinfold (approved):

provider-service/base/pkg/server/resource/experiment_test.go
```diff
@@ -0,0 +1,158 @@
...
+				Expect(err).ToNot(HaveOccurred())
+
+				expectedErr := errors.New("some-error")
+
+				mockProvider.On("CreateExperiment", ed).Return("", expectedErr)
+
+				response, err := exp.Create(jsonExperiment)
+
+				Expect(err).To(Equal(expectedErr))
+				Expect(response).To(Equal(ResponseBody{Id: ""}))
```
grahamia, line 64:
```suggestion
				Expect(response).To(Equal(ResponseBody{}))
```
is that the same 🤔

provider-service/base/pkg/server/resource/experiment_test.go
```diff
@@ -0,0 +1,158 @@
...
+				jsonExperiment, err := json.Marshal(ed)
+				Expect(err).ToNot(HaveOccurred())
+
+				expectedErr := errors.New("some-error")
+				id := "some-id"
+				mockProvider.On("UpdateExperiment", ed, id).Return("", expectedErr)
+				resp, err := exp.Update(id, jsonExperiment)
+
+				Expect(err).To(Equal(expectedErr))
+				Expect(resp.Id).To(BeEmpty())
```
grahamia, line 108: Same as previous comment

provider-service/base/pkg/server/resource/experiment_test.go
```diff
@@ -0,0 +1,158 @@
...
+			It("returns the id of the resource", func() {
+				ed := ExperimentDefinition{}
+				jsonExperiment, err := json.Marshal(ed)
+				Expect(err).ToNot(HaveOccurred())
+
+				id := "some-id"
+
+				mockProvider.On("CreateExperiment", ed).Return(id, nil)
+				response, err := exp.Create(jsonExperiment)
+				Expect(err).ToNot(HaveOccurred())
```
grahamia, line 35:
```suggestion

				Expect(err).ToNot(HaveOccurred())
```

provider-service/base/pkg/server/resource/experiment_test.go
```diff
@@ -0,0 +1,158 @@
...
+		})
+	})
+
+	Context("Delete", func() {
+		When("valid id is passed and provider operations succeed", func() {
+			It("return no error", func() {
+				id := "some-id"
+				mockProvider.On("DeleteExperiment", id).Return(nil)
+				err := exp.Delete(id)
+				Expect(err).ToNot(HaveOccurred())
```
grahamia, line 119:
```suggestion

				Expect(err).ToNot(HaveOccurred())
```

I actually prefer a space on both sides of the actual thing under test but dont think we have a standard

provider-service/base/pkg/server/resource/experiment_test.go
```diff
@@ -0,0 +1,158 @@
...
+		})
+
+		When("provider errors", func() {
+			It("errors", func() {
+				id := "some-id"
+				expectedErr := errors.New("some-error")
+				mockProvider.On("DeleteExperiment", id).Return(expectedErr)
+
+				err := exp.Delete(id)
+				Expect(err).To(Equal(expectedErr))
```
grahamia, line 130:
```suggestion

				Expect(err).To(Equal(expectedErr))
```

provider-service/base/pkg/server/resource/provider.go
```diff
@@ -0,0 +1,83 @@
...
+	TfxComponents string                `json:"tfxComponents" yaml:"tfxComponents"`
+	Env           []apis.NamedValue     `json:"env" yaml:"env"`
+	BeamArgs      []apis.NamedValue     `json:"beamArgs" yaml:"beamArgs"`
+}
+
+// Manifest represents the output of the python compile step, and describes what
+// vertex ai or kubeflow pipelines should do.
+type PipelineDefinitionWrapper struct {
+	PipelineDefinition PipelineDefinition `json:"pipelineDefinition"`
+	Manifest           json.RawMessage    `json:"manifest,omitempty"`
```
grahamia, line 23: I really dont like manifest here, doesnt explain what it is at all, hence your added comment, would CompiledPipeline or something similar to that make more sense?

provider-service/base/pkg/server/server.go
```diff
@@ -0,0 +1,141 @@
...
+	return func(w http.ResponseWriter, r *http.Request) {
+		id := chi.URLParam(r, "id")
+		decodedId, err := url.PathUnescape(id)
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
+		}
+		body, err := io.ReadAll(r.Body)
+		if err != nil {
+			http.Error(w, "Failed to read request body", http.StatusBadRequest)
```
grahamia, line 63: Does an error returned by reader mean its a bad request? 🤔

provider-service/base/pkg/server/server.go
```diff
@@ -0,0 +1,141 @@
...
+			return
+		}
+		defer r.Body.Close()
+
+		resp, err := hr.Update(decodedId, body)
+		if err != nil {
```
grahamia, line 69: I think the type of the error will have to be checked here as the error could be a 400 or a 422 here (but guess with 422 we would have to have strict json validation going on which dont think is a go thing

provider-service/base/pkg/server/resource/experiment_test.go
chr12c: yea it's the same. I've decided to go with always asserting on the entire response rather than just the id in case changes are made the the body in the future, then the test will break if someone forgets to update the test.
Line before/after each assertion block: https://github.com/sky-uk/kfp-operator/pull/435

provider-service/base/pkg/server/resource/experiment_test.go
```diff
@@ -0,0 +1,158 @@
...
+			})
+		})
+
+		When("invalid json is passed", func() {
+			It("errors", func() {
+				invalidJson := []byte(`/n`)
+				resp, err := exp.Update("some-id", invalidJson)
+
+				Expect(err).To(HaveOccurred())
+				Expect(resp.Id).To(BeEmpty())
```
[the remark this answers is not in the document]
chr12c: https://github.com/sky-uk/kfp-operator/pull/435

provider-service/base/pkg/server/resource/experiment_test.go
chr12c: https://github.com/sky-uk/kfp-operator/pull/435

provider-service/base/pkg/server/resource/experiment_test.go
chr12c: https://github.com/sky-uk/kfp-operator/pull/435

provider-service/base/pkg/server/resource/experiment_test.go
chr12c: https://github.com/sky-uk/kfp-operator/pull/435

before/after each assertion block rather than every assertion looks decent to me

provider-service/base/pkg/server/resource/experiment_test.go
chr12c: https://github.com/sky-uk/kfp-operator/pull/435

provider-service/base/pkg/server/server.go
chr12c: Yea fair point. We rely a lot on built-in errors, but now that we have a http server (albeit internally), to make some of the status codes actually make sense we have to start leaning on some domain specific errors, otherwise we’ll be returning 5xx for a lot of things.
Will potentially need errors specific to each layer of abstraction, then at some point pattern match (or whatever the golang equivalent is) against it to return the correct error.

chr12c: https://github.com/sky-uk/kfp-operator/pull/435/commits/8417f5dd922db6ad58714640f012ce8cb41b4a6a

provider-service/base/pkg/server/server.go
chr12c: should be 500
https://github.com/sky-uk/kfp-operator/pull/435/commits/d551d41a307b061d39c68e46c66f2d85f96efad5

provider-service/base/pkg/server/resource/provider.go
chr12c: are you referring to the name `Manifest` or the name `PipelineDefinitionWrapper`?

grahamia: the name Manifest, it means nothing to me 🤷🏼

chr12c: https://github.com/sky-uk/kfp-operator/pull/435/commits/fd433954e9d42753fac7a71434124b7979494d0e

chr12c: https://github.com/sky-uk/kfp-operator/pull/435/commits/d60451bde95615c8ec0c422950f1ce7c4736060c

grahamia (approved):
LGTM

PR #445, opened 2025-01-22
Track Provider resource state

### LINKED ISSUE #443: Track Provider resource state
## Overview

The KFP Operator tracks the state of custom resources using a `SynchronizationState` field. It also sets a condition (`SynchronizationSucceeded`) that aligns with that field and adds K8s events (as outlined in #74).

We want to do the same for Provider resources so they aligns with our other custom resources. In the case of Providers, the `SynchronizationState` will track the state of the Deployment and Service that the provider controller manages.

### Possible SynchronizationState values

- `Creating`: If the Deployment or Service don't exist (in which case the controller will create them)
- `Updating`: If the Deployment or Service exist but don't match the desired state described on the Provider resource (in which case the controller will update them to match)
- `Deleting`: If the Provider resource has been deleted (in which case the controller will delete the Deployment and Service)
- `Succeeded`: If the Provider resource reconciliation has completed with no failures
- `Failed`: If the Provider resource reconciliation has failed for any reason
- `Deleted`: This is used on custom resources that depend on a Provider to exist (e.g. Pipelines and RunConfigurations). It is used in any of these cases:
    - The resource has been successfully deleted (test If the Provider resource reconciliation has completed with no failures) - Note: This is a bit confusing - why (and how?) is the state set to Deleted when the resource has been deleted? 
    - A deletion for a resource has been requested, but there is no provider ID set on the resource, and therefore the deletion cannot be carried out (test [here](https://github.com/sky-uk/kfp-operator/blob/master/controllers/pipelines/state_handler_test.go#L529-L535) and [here](https://github.com/sky-uk/kfp-operator/blob/master/controllers/pipelines/state_handler_test.go#L547-L553)) - This case doesn't apply to Provider resources

## Acceptance Criteria

- [ ] The Provider controller tracks the state of provider resources. The state consists of:
    - `SynchronizationState`
    - `Synchronized` condition
    - Events (as outlined in #74 and implemented for our other CRDs)

alexgeorgousis:
Closes #443 

## Tasks

- [ ] Set `SynchronizationState` on `Provider` resource
    - [ ] Write decoupled tests for all state transitions
    - [ ] Current problematic scenario: failing deployment and then fixing it (add a test case for it)
    - [ ] Case to cover: What if Creating -> Deployment create operation fails -> Failed -> Creating -> ... forever
    - [ ] We don't cover failure scenarios in the pipeline decoupled tests - probably want some right?
- [ ] Set `Synchronized` condition (we do this for all other resources)???
- [ ] What about `ObservedGeneration`? (It's on the CRD but we never set it explicitly)
- [ ] Set events on provider resource
- [ ] Add/update integration tests if needed


alexgeorgousis:
Closing in favour of the work on [track-provider-synchronisation-state](https://github.com/sky-uk/kfp-operator/compare/master...track-provider-synchronisation-state) branch.

PR #453, opened 2025-01-24
Experiment service debugging

chr12c:
Closes # .

## Tasks

- [ ] ...
- [ ] Documentation updated


chr12c:
it was accidentally setting the call options to nil causing the panic

PR #472, opened 2025-02-06
Improve Makefiles

chr12c:
- The `website` target command doesn't specify a target for `docs-gen`, which means it will default to the first target in `docs-gen/Makefile`. This happens to be `build` (which is what we want), but could change in the future by accident so it's better to be specific.

## Tasks
- [x] Specify build target for `docs-gen` in root `Makefile`


chr12c:
Closing because there are new changes to the Makefiles

Issue #464, opened 2025-01-30
Call provider API from controller workflows


stefania-popescu:
## Overview

Currently, the argo workflows that manage CRUD operations for resources call the provider CLI during the `provider` step. Now it needs to be changed so that the workflow calls the `provider-service` API. Each CLI call should have a corresponding API call that needs to be configured.

There's an integration test that stubs out the (old) provider interface; when the `provider-service` API is called, the new provider interface will need to be stubbed out since the shape is slightly different.

## Tasks
- [x] Create stubbed provider service for testing
- [x] Create new template and step to combine `resource.json` and `resource-definition.yaml`
- [ ] Change `provider` step in argo workflows to call `provider-service` API
- [ ] Remove `Image` field from Provider CRD and everywhere it's used in workflows (we don't need a different image per provider, just pass diff parameters to API call)

chr12c:
To call the provider service api, we need to be able to send a URL encoded id from the workflow. There’s 2 options:
- encode the URL from the workflow
- encode the URL from the Go code.

I think most would agree that leaking business logic into a yaml file isn't ideal since it's easier to control and unit test if it was within the Go code. To do this we can initially introduce another input parameter into the  `delete`-related workflow templates; this won't break anything because nobody will be using it.

alexgeorgousis:
> To call the provider service api, we need to be able to send a URL encoded id from the workflow. The delete api in particular expects an encoded id since they could involve `/`.
> 
> There’s 2 options:
> 
> * encode the URL from the workflow
> * encode the URL from the Go code.
> 
> I think most would agree that leaking business logic into a yaml file isn't ideal since it's easier to control and unit test if it was within the Go code. To do this we can initially introduce another input parameter into the `delete`-related workflow templates; this won't break anything because nobody will be using it.

What would the input parameter be? Also, is this another reason to keep the provider CLI and make the API request from there? Along with everything else we'll need to handle. Basically what I'm saying is:

> I think most would agree that leaking business logic into a yaml file isn't ideal

Can we use this as an argument to have the entire API request logic happen inside the CLI?

chr12c:
The parameter is the encoded resource-id.

I don't think this is a reason to keep the CLI just because the CLI can technically make an API request. This doesn't count as a reason for keeping it if the entire CLI can be removed if the template can make the call. The logic of encoding the id can be completely encapsulated in the construction of the workflow itself, i.e. the workflow factory. This doesn't seem too difficult.

[...]

PR #473, opened 2025-02-06
Stub provider service

chr12c:
Related to #464
Stub provider service that returns canned responses which are exposed in a new common module for the workflow integration tests to access.

## Tasks

- [x] Implement Stub provider service
- [x] Tested manually via docker


provider-service/stub/internal/provider/provider.go
```diff
@@ -0,0 +1,122 @@
...
+}
+
+func New(logger logr.Logger) resource.Provider {
+	return &StubProvider{logger}
+}
+
+func (p *StubProvider) CreatePipeline(
+	pdw resource.PipelineDefinitionWrapper,
+) (string, error) {
+	if strings.EqualFold(pdw.PipelineDefinition.Name.Name, "create-pipeline-fail") {
```
TobyPinfold, line 22: Can we move all these matcher ids to be constants in a shared package somewhere, that way the caller / test suite can guarantee they match up correctly.

provider-service/stub/internal/config/config.go
```diff
@@ -0,0 +1,44 @@
...
+
+	var config Config
+	if err := viper.ReadInConfig(); err != nil {
+		return nil, fmt.Errorf("fatal error reading config: %w", err)
+	}
+	if err := viper.Unmarshal(&config); err != nil {
+		return nil, fmt.Errorf("fatal error unmarshalling config %w", err)
+	}
+	return &config, nil
+}
```
alexgeorgousis, line 44: Could all be in 1 func like in https://github.com/sky-uk/kfp-operator/blob/master/provider-service/base/pkg/config/config.go#L28-L49

It's okay either way though if you feel strongly about this 😄

provider-service/stub/internal/provider/provider.go
chr12c: shared package outside of `provider-service/stub`?

provider-service/stub/internal/config/config.go
chr12c: Why is the generic one `json` 🤣

provider-service/stub/internal/provider/provider.go
TobyPinfold: if possible, we would likely need it within the argo workflow test suite

provider-service/stub/internal/config/config.go
TobyPinfold: can try removing the `SetConfigType` entirely and let Viper determine the type it should use based on the extension of `config`

alexgeorgousis: That'd be ideal 🪄

Wanna try that in the main `LoadConfig` func as well @chr12c? Otherwise can do it in separate small PR.

chr12c: commit to use the shared config loader https://github.com/sky-uk/kfp-operator/pull/473/commits/8ae2de88d286710b0d61488e17e9da546953045c

chr12c: > can try removing the `SetConfigType` entirely and let Viper determine the type it should use based on the extension of `config`

No, the type has to remain `json`, otherwise viper doesn't know how to unmarshal. I guess we're at crossroads now:
- leave in the `json` despite never actually loading in any config. It's slightly confusing but allows us to keep using a generic config loader
- go back to a non-generic config loading method and set the defaults in an external file that gets loaded int. Now `json` (or `yaml`) being present in the code is less confusing

chr12c: @TobyPinfold https://github.com/sky-uk/kfp-operator/pull/473/commits/bc99adc0eb15dac73a44af8ae4bc9da0657fa4fd

common module

provider-service/stub/internal/provider/provider.go
chr12c: https://github.com/sky-uk/kfp-operator/pull/473/commits/bc99adc0eb15dac73a44af8ae4bc9da0657fa4fd

alexgeorgousis (approved):

provider-service/stub/internal/provider/provider.go
TobyPinfold: missing the actual matcher, would expect `"create-pipeline-fail"` to be exported for use within the test suite, so it can ensure the pipeline it calls the stub with will match what is expected

chr12c: 8d3f4d77f01908a3687c8bb2c408b9b03167e976

aidandunlop (approved):

grahamia (approved):

PR #474, opened 2025-02-07
Pass all config to provider service from controller

### LINKED ISSUE #452: Pass all config to provider service from controller
## Overview

Currently, this is how the provider service receives its config:
1. Get _some_ of the config the service needs from the Provider resource and the global KFP Operator config, and set it on the provider service Deployment when creating it in the Provider controller
2. When creating the provider service Deployment, pass the provider name to the pod as an env var (`PROVIDERNAME`)
3. In the provider service code, use that env var to fetch the Provider resource and load the rest of the config the service needs

Steps (2) and (3) made sense when there was no Provider controller to pass the config to the provider service, but now that there is, we can pass all the config the service needs in step (1) and drop everything else.

## Acceptance Criteria

- [x] All the config the provider service needs is set on the Deployment by the Provider controller
- [x] There is no provider-loading logic in the provider service code

alexgeorgousis:
Closes #452

## Overview

As explained in the related issue, we currently set some service config in the controller and some at service startup by loading the `Provider` resource (which is unnecessary).

This PR:
- Passes all config the service needs as env vars on the service `Deployment`
- Removes all `Provider`-loading logic from the service (both kfp and vai) and just reads all config from env vars

## Tasks

- [x] Load all config from env vars in kfp
- [x] Load all config from env vars in vai
- [x] Refactor vai/cmd/main.go and kfp/cmd/main.go to be consistent with each other
- [x] Refactor: have 1 generic `LoadConfig[T any]` function
- [x] Read service port in manager configmap from helm values instead of hardcoding


alexgeorgousis:
Current problem testing kfp provider in sandbox:

```
{"error":"Service \"provider-kfp-4w4lp\" is invalid: [spec.ports[0].port: Invalid value: 0: must be between 1 and 65535, inclusive}
```

As far as I can tell we set the port to 8080 🤷‍♂️

apis/pipelines/v1alpha6/provider_types.go
```diff
@@ -33,10 +32,10 @@ type ProviderSpec struct {
...
 	ExecutionMode string `json:"executionMode" yaml:"executionMode"`
 	// +kubebuilder:validation:Pattern:=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
-	ServiceAccount      string                           `json:"serviceAccount" yaml:"serviceAccount"`
-	DefaultBeamArgs     []apis.NamedValue                `json:"defaultBeamArgs,omitempty" yaml:"defaultBeamArgs,omitempty"`
-	PipelineRootStorage string                           `json:"pipelineRootStorage" yaml:"pipelineRootStorage"`
-	Parameters          map[string]*apiextensionsv1.JSON `json:"parameters,omitempty" yaml:"parameters,omitempty"`
+	ServiceAccount      string            `json:"serviceAccount" yaml:"serviceAccount"`
+	DefaultBeamArgs     []apis.NamedValue `json:"defaultBeamArgs,omitempty" yaml:"defaultBeamArgs,omitempty"`
+	PipelineRootStorage string            `json:"pipelineRootStorage" yaml:"pipelineRootStorage"`
+	Parameters          map[string]string `json:"parameters,omitempty" yaml:"parameters,omitempty"`
```
alexgeorgousis, line 38: We updated the type of `Provider.Spec.Parameters` [here](https://github.com/sky-uk/kfp-operator/pull/353/commits/cacab20dbe85f1f7ffdceccc1fc53ea9ea770f8b) from `map[string]string` to map[string]*JSON, but I'm not sure why.

- All parameters for the 2 providers we support are currently strings.
- Even if we did this to allow users to add params which are more complex structures without having to make code changes, that's not the case because we'd need to update the provider service code (where we redefine this `Parameters` struct)

alexgeorgousis, line 38: I assume the reason why we set the type of `Provider.Spec.Parameters` to `map[string]*apiextensionsv1.JSON` instead of `map[string]string` is because we support `parameters.maxConcurrentRunCount` (which is an `int`) in the VAI provider.

The problem is we want to pass all these `parameters` as environment variables to the provider service, so they need to be converted into strings.

It doesn't seem to be possible to do this:
1. Unmarshal the JSON parameter into `any` (because we don't know what the underlying type is)
2. Convert to string (this is where it fails because it needs to know the type)

So I propose we make parameters strings on the controller side so they can be set as env vars, and then deserialize them into their specific types in the provider service (which we already do).

provider-service/vai/cmd/main.go
```diff
@@ -26,57 +27,43 @@ func main() {
...
 	if err != nil {
+		logger.Error(err, "failed to load provider config", "provider", baseConfig.ProviderName, "namespace", baseConfig.Pod.Namespace)
 		panic(err)
 	}
+	logger.Info(fmt.Sprintf("loaded VAI provider config: %+v", vaiProviderConfig))
 
-	vaiConfig := vaiConfig.VAIProviderConfig{
-		Name: config.ProviderName,
-	}
+	RunServer(ctx, vaiProviderConfig, baseConfig)
```
alexgeorgousis, line 46: In vai main.go we just do `server.Start()`, but in kfp main.go we run it as a go routine. Is there a reason for that or just oversight?

controllers/pipelines/provider_controller.go
```diff
@@ -248,20 +251,45 @@ func populateServiceContainer(serviceContainerName string, podTemplate v1.PodTem
 		return nil, fmt.Errorf("unable to populate service container: container with name %s not found on deployment", serviceContainerName)
 	}
 
+	var paramsAsEnvVars []v1.EnvVar
+	for name, value := range provider.Spec.Parameters {
+		paramsAsEnvVars = append(paramsAsEnvVars, v1.EnvVar{Name: fmt.Sprintf("PARAMETERS_%s", strings.ToUpper(name)), Value: jsonToString(value)})
+	}
```
alexgeorgousis, line 257: The reason we have to do all this JSON parsing is because `Provider.Spec.Parameters` is of type `map[string]*JSON` instead of simply `map[string]string`. Is there a reason for this?

- Currently, the only non-string parameter we support is `maxConcurrentRunCount` in the VAI provider, which is an `int`. But we can easily make that a string on the Provider CRD and then parse it into an `int` in the provider service
- Judging from the tests, we're expecting complex JSON structures as provider parameters, but why? It seems a bit excessive if it's simply for "maybe in the future we'll need to support complex JSON params", doesn't it?

provider-service/base/pkg/config/config.go
```diff
@@ -27,28 +25,25 @@ type Server struct {
 	Port int    `mapstructure:"port"`
 }
 
-func LoadConfig(ctx context.Context) (*Config, error) {
-	logger := common.LoggerFromContext(ctx)
```
alexgeorgousis, line 31: I moved the logging out of the `LoadConfig` function to `main.go` (for each provider). Partly to keep this as a business-logic-only function and partly because we previously had some config-related logging in `main.go` and some in here (inconsistent).

provider-service/vai/cmd/main.go
```diff
@@ -26,66 +27,72 @@ func main() {
...
 	if err != nil {
+		logger.Error(err, "failed to load provider config", "provider", serviceConfig.ProviderName, "namespace", serviceConfig.Pod.Namespace)
 		panic(err)
 	}
+	logger.Info(fmt.Sprintf("loaded provider config: %+v", vaiProviderConfig), "provider", serviceConfig.ProviderName, "namespace", serviceConfig.Pod.Namespace)
 
-	vaiConfig := vaiConfig.VAIProviderConfig{
-		Name: config.ProviderName,
-	}
+	go runServer(ctx, vaiProviderConfig, serviceConfig)
```
alexgeorgousis, line 51: I moved the code that does `server.Start()` in a function (because it's identical to the kfp code and that's what we do there), but `servcer.Start()` wasn't running as a goroutine before, even though it is in kfp.

So I made it run in a goroutine, but is there a reason not to?

controllers/pipelines/provider_controller.go
grahamia: Provider specific parameters could be anything in the future, I would say it is better to handle that likely complexity from the start rather than having to shoehorn it in at a later date making bigger changes? Only allowing single value parameters would be a big restriction on the provider

controllers/pipelines/provider_controller_test.go
```diff
@@ -282,4 +285,32 @@ var _ = Context("Provider Controller", func() {
 
 		})
 	})
+
+	var _ = Describe("jsonToSTring", func() {
```
TobyPinfold, line 289:
```suggestion
	var _ = Describe("jsonToString", func() {
```

provider-service/vai/cmd/main.go
TobyPinfold: prevents any server blow up causing the eventing to also be blocked from running

controllers/pipelines/provider_controller_test.go
alexgeorgousis: Done in 688b5b1

TobyPinfold (approved):

controllers/pipelines/provider_controller.go
```diff
@@ -248,20 +251,39 @@ func populateServiceContainer(serviceContainerName string, podTemplate v1.PodTem
 		return nil, fmt.Errorf("unable to populate service container: container with name %s not found on deployment", serviceContainerName)
 	}
 
+	var paramsAsEnvVars []v1.EnvVar
```
TobyPinfold, line 254:
```suggestion
	paramsAsEnvVars := []v1.EnvVar{}
```

alexgeorgousis: Done in 12dd027 (fixed the length as well)

provider-service/vai/internal/provider/provider.go
```diff
@@ -20,17 +20,17 @@ import (
...
+	config         *config.VAIProviderConfig
 	fileHandler    FileHandler
 	pipelineClient client.PipelineJobClient
 	scheduleClient client.ScheduleClient
 	jobBuilder     JobBuilder
 	jobEnricher    JobEnricher
 }
 
-func NewProvider(
+func NewVAIProvider(
```
chr12c, line 31: there's no other provider in this context (vai), so NewProvider must imply vai

provider-service/base/pkg/config/config.go
```diff
@@ -57,16 +52,10 @@ func load() (*Config, error) {
...
 // this initialises the config in Viper with default empty values so that they can be overridden with env vars
-func initConfig() error {
-	config := Config{
-		Server: Server{
-			Host: "0.0.0.0",
-			Port: 8080,
-		},
-	}
+func initaliseConfig[T any](initConfig T) error {
 	viper.SetConfigType("json")
```
chr12c, line 56: method and description don't match up anymore.
`T` isn't really `any` is it? It's actually a requirement for it to have a `Server`

provider-service/vai/internal/provider/provider.go
alexgeorgousis: Yep thought about that too - at the end of the day I just wanted to make it consistent with KFP (where it's called `NewKfpProvider`), so I'm happy to change both 👍

provider-service/base/pkg/config/config.go
alexgeorgousis: It's actually used to load config into the provider-specific config structs in vai and kfp, not just the base config which includes `Server`.

So `initConfig()` takes a default/initial value for whatever config of type `T` you want to initialise.

provider-service/vai/internal/provider/provider.go
chr12c: `NewProvider` is the best we can do, given there are other services in the same `provider` package.

provider-service/base/pkg/config/config.go
chr12c: `T` must be constrained to the type `Server`, otherwise the provider service won't have a host/port to bind to.

provider-service/vai/internal/provider/provider.go
TobyPinfold: I think its best to just be NewKFPProvider and NewVAIProvider, given it matches the struct it creates, avoids any potential confusion as else you are relying on the named import being set correctly.

chr12c (approved):

PR #476, opened 2025-02-11
Track provider synchronisation state

TobyPinfold:
Closes #445 
 
 **Adds:**
  - Updates provider resources with state (generation, success / failed / updating) 
  - Adds test suites for provider controller including Reconcile method checks
  - Adds predicate filters to prevent repeat reconciliation requests created by actions performed by the controller.
 
 **Updates:**
  - Refactors the provider controller structure to improve readability, structure and test coverage.  
  - Moves logic out of provider controller into specialised managers for each sub-resource created by the controller.

**Fixes:**
 - Loading parameters into deployment spec did not maintain or respect order so un-neccesary updates were made to deployments even though their definitions were the same but in differing orders.

**Removes:**
 - Removes use of hashing annotations on provider sub-resources (service / deployment) as it provides no guarantee of the actual state of a resource once deployed to k8s where managed fields and cluster specific information is added.


controllers/pipelines/internal/provider/predicates/deployment_predicate.go
```diff
@@ -0,0 +1,25 @@
...
+	v1 "k8s.io/api/apps/v1"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+type DeploymentChangedPredicate struct {
+	predicate.Funcs
+}
+
+func (DeploymentChangedPredicate) Create(e event.CreateEvent) bool {
```
TobyPinfold, line 13: arguable wether this is still needed same for services but will check, though rest of PR is still reviewable

controllers/pipelines/internal/provider/predicates/service_predicate.go
```diff
@@ -0,0 +1,26 @@
...
+	if !ok {
+		return true
+	}
+
+	//Ignore create events caused by controller ops
+	if svc.Annotations != nil && svc.Annotations[ControllerManagedKey] == "true" {
+		return false
+	}
+
+	return true
```
peter-hazell, line 25: can be simplified same as above

controllers/pipelines/internal/testutil/testutil.go
```diff
@@ -16,3 +19,14 @@ var (
 	TestConfig config.KfpControllerConfigSpec
 	Provider   *pipelinesv1.Provider
 )
+
+func SchemeWithCRDs() *runtime.Scheme {
```
peter-hazell, line 23:
```suggestion
func SchemeWithCrds() *runtime.Scheme {
```

not overly precious about it but I prefer strict pascal case even using acronyms

controllers/pipelines/provider_deployment_manager.go
```diff
@@ -0,0 +1,178 @@
...
+
+func populateServiceContainer(serviceContainerName string, podTemplate corev1.PodTemplateSpec, provider *pipelinesv1.Provider) (*corev1.PodTemplateSpec, error) {
+	if !Exists(podTemplate.Spec.Containers, func(c corev1.Container) bool {
+		return c.Name == serviceContainerName
+	}) {
+		return nil, fmt.Errorf("unable to populate service container: container with name %s not found on deployment", serviceContainerName)
+	}
+
+	envVars := []corev1.EnvVar{{
+		Name:  "PROVIDERNAME",
```
peter-hazell, line 147: Wonder if we should put these env vars somewhere common also?

controllers/pipelines/internal/provider/predicates/deployment_predicate.go
aidandunlop: can we make this generic for both Deployments and Services? Looks the same except the resource type?

controllers/pipelines/provider_status_manager_test.go
```diff
@@ -0,0 +1,73 @@
...
+			Expect(provider.Status.SynchronizationState).To(Equal(expectedStatus))
+			Expect(provider.Status.Conditions[0].Message).To(Equal(expectedMessage))
+			Expect(provider.Status.ObservedGeneration).To(Equal(provider.Generation))
+		})
+
+		It("should return error if the status is not updated", func() {
+			// No provider is created so no status to update
+			err := statusManager.UpdateProviderStatus(ctx, provider, apis.Succeeded, "test")
+			Expect(err).To(HaveOccurred())
+		})
```
aidandunlop, line 71: could add a test for when we updated the provider status with a non-succeeded state? To check that the ObservedGeneration _doesn't_ get updated

controllers/pipelines/internal/testutil/testutil.go
```diff
@@ -16,3 +19,14 @@ var (
...
+	scheme := runtime.NewScheme()
+
+	groupVersion := schema.GroupVersion{Group: "pipelines.kubeflow.org", Version: "v1alpha6"}
```
grahamia, line 26: Is groupVersion not accessible in apis/pipelines/v1alpha6?

controllers/pipelines/provider_controller.go
```diff
@@ -54,347 +81,92 @@ func NewProviderReconciler(ec K8sExecutionContext, config config.KfpControllerCo
...
 
-		if deploymentIsOutOfSync(existingDeployment, desiredDeployment) {
-			logger.V(2).Info("resource hash mismatch, updating deployment")
-			existingDeployment = syncDeployment(existingDeployment, desiredDeployment)
-			if err = r.EC.Client.Update(ctx, existingDeployment); err != nil {
-				logger.Error(err, "unable to update provider deployment", "deployment", desiredDeployment)
-				return err
+			if err := r.DeploymentManager.Create(ctx, desiredDeployment, &provider); err != nil {
+				if err2 := r.StatusManager.UpdateProviderStatus(ctx, &provider, apis.Failed, "Failed to reconcile subresource deployment"); err2 != nil {
+					err = errors.Join(err, err2)
```
grahamia, line 113: Shall we change all these to deployErr and statusErr or something rather than err2?

controllers/pipelines/provider_deployment_manager_test.go
```diff
@@ -0,0 +1,316 @@
...
+						},
+					},
+				},
+			},
+		}
+	})
+
+	var _ = Describe("Create", func() {
+
+		Specify("Should return nil if the deployment is created successfully", func() {
```
grahamia, line 69: I would prefer these to be worded more along the lines of should return no error rather than return nil?

controllers/pipelines/provider_deployment_manager_test.go
```diff
@@ -0,0 +1,316 @@
...
+			Expect(err).ToNot(HaveOccurred())
+
+			result, err := deploymentManager.Get(context.Background(), provider)
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(result.Name).To(Equal(deployment.Name))
+			Expect(result.Namespace).To(Equal(deployment.Namespace))
+		})
+
+		It("should return an NotFound error if the deployment does not exist", func() {
```
grahamia, line 171:
```suggestion
		It("should return a NotFound error if the deployment does not exist", func() {
```

controllers/pipelines/internal/provider/predicates/deployment_predicate.go
TobyPinfold: I'm hopefully binning this part off anyway but if it is still required, yes 👍

controllers/pipelines/internal/testutil/testutil.go
TobyPinfold: yes and is now used in [a2f6bc4](https://github.com/sky-uk/kfp-operator/pull/476/commits/a2f6bc4c820a0838497ec6c03768d77762610472)

controllers/pipelines/provider_controller.go
TobyPinfold: done in [a2f6bc4](https://github.com/sky-uk/kfp-operator/pull/476/commits/a2f6bc4c820a0838497ec6c03768d77762610472) 
kept err as default but the err2 from status is now statusErr which is more consistent with other code

controllers/pipelines/provider_deployment_manager_test.go
TobyPinfold: done in [a2f6bc4](https://github.com/sky-uk/kfp-operator/pull/476/commits/a2f6bc4c820a0838497ec6c03768d77762610472)

controllers/pipelines/provider_deployment_manager_test.go
TobyPinfold: done in [a2f6bc4](https://github.com/sky-uk/kfp-operator/pull/476/commits/a2f6bc4c820a0838497ec6c03768d77762610472)

controllers/pipelines/internal/testutil/mock_service_manager.go
```diff
@@ -0,0 +1,50 @@
...
+	args := m.Called(a, b)
+	if args.Get(0) != nil {
+		return args.Get(0).(bool)
+	}
+	panic("mock not set")
+}
+
+func (m *MockServiceManager) Construct(provider *pipelinesv1.Provider) *corev1.Service {
+	args := m.Called(provider)
+
```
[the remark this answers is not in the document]
TobyPinfold: done in [a2f6bc4](https://github.com/sky-uk/kfp-operator/pull/476/commits/a2f6bc4c820a0838497ec6c03768d77762610472)

controllers/pipelines/internal/provider/predicates/deployment_predicate.go
TobyPinfold: removed this code block within [a2f6bc4](https://github.com/sky-uk/kfp-operator/pull/476/commits/a2f6bc4c820a0838497ec6c03768d77762610472)

controllers/pipelines/internal/provider/predicates/deployment_predicate.go
```diff
@@ -0,0 +1,25 @@
...
+	if !ok {
+		return true
+	}
+
+	//Ignore create events caused by controller ops
+	if svc.Annotations != nil && svc.Annotations[ControllerManagedKey] == "true" {
+		return false
+	}
+
+	return true
```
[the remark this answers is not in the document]
TobyPinfold: removed this code block within [a2f6bc4](https://github.com/sky-uk/kfp-operator/pull/476/commits/a2f6bc4c820a0838497ec6c03768d77762610472)

controllers/pipelines/internal/provider/predicates/service_predicate.go
TobyPinfold: removed this code block within [a2f6bc4](https://github.com/sky-uk/kfp-operator/pull/476/commits/a2f6bc4c820a0838497ec6c03768d77762610472)

controllers/pipelines/provider_status_manager_test.go
TobyPinfold: added in [cd86f99](https://github.com/sky-uk/kfp-operator/pull/476/commits/cd86f990f90fd194eb9cb15c5ae8f88051f1bc10)

controllers/pipelines/provider_deployment_manager_test.go
aidandunlop: this one still looks the same?

TobyPinfold: cmd shift r has failed me

TobyPinfold: actually fixed in [dd93fa3](https://github.com/sky-uk/kfp-operator/pull/476/commits/dd93fa33b201c6b8ef4c023e01b2289758bc8680)

aidandunlop (approved):

controllers/pipelines/provider_controller.go
```diff
@@ -54,347 +79,92 @@ func NewProviderReconciler(ec K8sExecutionContext, config config.KfpControllerCo
...
-	err := r.EC.Client.NonCached.List(ctx, sl, &client.ListOptions{
-		Namespace: provider.Namespace,
-	})
-	if err != nil {
-		return nil, err
-	}
-	for _, svc := range sl.Items {
-		if metav1.IsControlledBy(&svc, &provider) {
-			return &svc, nil
+			// delete, to allow a new one to be created avoiding issues with immutability
```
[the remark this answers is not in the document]
TobyPinfold: 99% of the time there will be no need to recreate the service, its only in the 1% of time where a field is changed from the manager config ie ports that we need to update the service. At which point the existing service would already be out of sync with the deployment and be broken anyhow.

Would suggest we instead make the client calling the provider service resilient to the seconds long break in servicing, the argo workflows can retry and backoff so would suggest we add that behaviour there to offset this.

controllers/pipelines/internal/testutil/mock_deployment_manager.go
```diff
@@ -0,0 +1,50 @@
+//go:build unit
+
+package testutil
```
chr12c, line 3: can you put all of the mocks in `internal/mocks` like in other places we use testify.

apis/pipelines/v1alpha6/provider_types.go
```diff
@@ -52,17 +52,30 @@ func (p *Provider) SetStatus(status Status) {
...
 		Namespace: p.Namespace,
 	}
 }
 
-func (e Provider) GetKind() string {
+func (p *Provider) GetKind() string {
 	return "provider"
 }
 
+func (p *Provider) StatusWithCondition(message string) Status {
```
chr12c, line 66: This function mutates the Status field of Provider, so what are you doing with the returned Status?

controllers/pipelines/internal/testutil/mock_deployment_manager.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+	return deployment, args.Error(1)
+}
+
+func (m *MockDeploymentManager) Equal(a, b *appsv1.Deployment) bool {
+	args := m.Called(a, b)
+	if args.Get(0) != nil {
+		return args.Get(0).(bool)
+	}
+	panic("mock not set")
```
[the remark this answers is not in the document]
TobyPinfold: Its either panic or you must return something in the function for the compiler to be happy, in this case it makes sense to panic instead of blindy returning a value for the bool, so it wont pass / fail a test silently.

apis/pipelines/v1alpha6/provider_types.go
TobyPinfold: Was a leftover, have removed return + type from function in  [f75f6b0](https://github.com/sky-uk/kfp-operator/pull/476/commits/f75f6b025e2bf8e3567c4827be009be352ef678a)

controllers/pipelines/internal/testutil/mock_deployment_manager.go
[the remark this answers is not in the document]
chr12c: I'm confused - this `Equal` method is surely called somewhere in the chain of logic in provider controller (unless I'm wrong, but if that's the case then why would this method exist in the first place), so if it should be called somewhere then the result should be mocked out when you call `.Return(true)` or `.Return(false)`

controllers/pipelines/provider_service_manager.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+func (sm ServiceManager) Get(ctx context.Context, owner *pipelinesv1.Provider) (*corev1.Service, error) {
+	sl := &corev1.ServiceList{}
+	err := sm.client.NonCached.List(ctx, sl, &client.ListOptions{
+		Namespace: owner.Namespace,
+	})
+	if err != nil {
+		return nil, err
+	}
```
chr12c, line 68: combine the error check with the call to `List`.

controllers/pipelines/provider_service_manager_test.go
```diff
@@ -0,0 +1,222 @@
...
+		It("should return a NotFound error if the service does not exist", func() {
+			service, err := serviceManager.Get(ctx, provider)
+			Expect(service).To(BeNil())
+			Expect(err).To(Equal(apierrors.NewNotFound(schema.GroupResource{}, "")))
+		})
+	})
+
+	var _ = Describe("Construct", func() {
+
+		It("should return a service if constructed correctly", func() {
```
chr12c, line 170: This method never errors so it return a ptr to a service regardless. So there is no such thing as "correctly"

controllers/pipelines/internal/testutil/mock_deployment_manager.go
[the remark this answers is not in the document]
TobyPinfold: It is not necessarily used in every situation as it can be short circuited by provider being out of sync and so enforce reconciliations.

But the point still stands,

The Equal method setup will respond with true or false, but I'm configuring it here, If its not been configured correctly I don't want to blindly respond with a true or false default value as that could incorrectly pass the test. 

IMO The above panic is correct, if its used and not been configured correctly then it should error

controllers/pipelines/provider_service_manager_test.go
TobyPinfold: left over it used to error, will update

controllers/pipelines/provider_service_manager.go
```diff
@@ -0,0 +1,107 @@
...
+	client *controllers.OptInClient
+	scheme *runtime.Scheme
+	config *config.KfpControllerConfigSpec
+}
+
+func (sm ServiceManager) Create(ctx context.Context, new *corev1.Service, provider *pipelinesv1.Provider) error {
+	logger := log.FromContext(ctx)
+
+	if err := ctrl.SetControllerReference(provider, new, sm.scheme); err != nil {
+		logger.Error(err, "unable to set controller reference on service")
```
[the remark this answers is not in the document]
TobyPinfold: done in [b568b7e](https://github.com/sky-uk/kfp-operator/pull/476/commits/b568b7e36ab9bfde8f7dc838e8d8585f865524c5)

controllers/pipelines/internal/testutil/mock_deployment_manager.go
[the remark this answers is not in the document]
chr12c: I still don't understand what you mean. If it's not been configured correctly then it should already error. Perhaps take a look at the mock run service:
https://github.com/sky-uk/kfp-operator/blob/0ddbfddb62eb77664548a73932cca28e9e4ce41a/provider-service/kfp/internal/mocks/run_service.go#L19

In your case you should be using a straight
```go
func (m *MockDeploymentManager) Equal(a, b *appsv1.Deployment) bool {
	args := m.Called(a, b)
	return args.Bool(0)
}
```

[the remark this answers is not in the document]
TobyPinfold: ahhh I didn't know there was a args.Bool func 🤦 , now I see what you meant fixed in [3e1cc73](https://github.com/sky-uk/kfp-operator/pull/476/commits/3e1cc73c3513100bac7ceb17d1f4e83c8ca191ea)

chr12c (approved):

aidandunlop (approved):

controllers/pipelines/provider_controller.go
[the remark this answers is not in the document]
TobyPinfold: Agreed on this elsewhere

Issue #376, opened 2024-10-16
Change from Provider CLI to Provider API


grahamia:
WIP

**Is your feature request related to a problem? Please describe.**
Currently monitoring of workflow failures is not done. So if a call to VAI for example fails then the workflow finishes and no one is the wiser.

Currently the various workflows KFP Operator triggers start a pod for the provider image which provides a CLI for the various bits of functionality.

e.g. `/provider --provider <config location> pipeline create --pipeline-definition /resource-definition.yaml`

To allow for better monitoring of provider events turning this step into calling an api on a running deployment that can expose metrics (using open telemetry) (rather than relying on publishing metrics to the argo workflow controller, which we have already experienced is not the best solution).

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

TobyPinfold:
We have come up with a draft proposal for the api spec for this service:

```yaml
openapi: 3.0.3
info:
  title: Provider Resource Management
  description: Provider API for managing resources.
  version: 0.0.1
paths:
  /resource/{type}:
    put:
      tags:
        - Resource Operations
      summary: Create a new resource
      description: Creates a resource using the provided JSON body to identify the resource by name.
      operationId: createResource
      requestBody:
        required: true
        content:
          application/json:
            schema:
              oneOf:
              - $ref: '#/components/schemas/PipelineResource'
              - $ref: '#/components/schemas/RunResource'
              - $ref: '#/components/schemas/ExperimentResource'
              - $ref: '#/components/schemas/RunScheduleResource'
      responses:
        '201':
          description: Resource created successfully.
        '400':
          description: Bad request, invalid input.
  /resource/{type}/{name}:
    post:
      tags:
        - Resource Operations
      summary: Update an existing resource
      description: Updates a resource using the provided JSON body and the name passed as a path parameter to identify the resource.
      operationId: updateResource
      parameters:
        - name: name
          in: path
          required: true
          schema:
            type: string
          description: The name of the resource to update.
      requestBody:
        required: true
        content:
          application/json:
            schema:
              oneOf:
              - $ref: '#/components/schemas/PipelineResource'
              - $ref: '#/components/schemas/RunResource'
              - $ref: '#/components/schemas/ExperimentResource'
              - $ref: '#/components/schemas/RunScheduleResource'
      responses:
        '200':
          description: Resource updated successfully.
        '400':
          description: Bad request, invalid input.
        '404':
          description: Resource not found.
    delete:
[...]

PR #477, opened 2025-02-12
Merge local minikube and integration tests env setups

### LINKED ISSUE #479: Merge local minikube and integration tests env setups
We currently have 2 minikube setups:
- One spun up by the `integration-test-up` make target used for workflow factory integration tests
- One spun up by the `minikube-up` make target used to provide a local manual testing environment

I think the 2 should be merged because:
1. Makefile cohesion - it's strange to have 2 separate things that do something quite similar
2. Provider tests: We should use 1 `integration-test-up` make target for all integration tests, but we need some things from `minikube-up` (e.g. installing the operator) for provider tests to work - so we might as well merge them instead of duplicating

alexgeorgousis:
Closes #479 

## Tasks

- [ ] Make `integration-test` work with `minikube-up` (see comments below for latest state)
	- [ ] Once it's all done, remove `integration-test-up`
- [ ] Make sure everything the `integration-test-down` Make target does is done by `minikube-down` and then remove it
- [ ] Documentation updated (dev guide?)


alexgeorgousis:
**Current problem**

While running `make integration-test`, workflows run but fail with:

```
Error (exit code 64): workflowtaskresults.argoproj.io is forbidden: User "system:serviceaccount:kfp-operator-system:kfp-operator-argo" cannot create resource "workflowtaskresults" in API group "argoproj.io" in the namespace "kfp-operator-system"
```

minikube.mk
```diff
@@ -1,7 +1,9 @@
+ARGO_WORKFLOWS_CHART_VERSION=0.17.0
+ARGO_EVENTS_CHART_VERSION=2.4.1
```
alexgeorgousis, line 2: In `integration-test-up` we use a lengthy `sed` command to grab the argo workflows version from `go.mod`. This doesn't work here because we're installing argo workflows using the helm chart, so we need the helm chart version (same goes for argo events).

There is a workaround (haven't tried it) but it's a bit hacky: we can override the image versions in the helm values to point to the correct version of argo workflows (same with events). It would look something like this:

```
$(HELM) install argo-workflows argo/argo-workflows -n argo --create-namespace \
  --set "argo-workflows.controller.image.tag=3.4.5" \
  --set "argo-workflows.server.image.tag=3.4.5"
```

alexgeorgousis:
### Current problem

The integration tests run and create workflows successfully, but the workflows fail at the `provider` step with:

```
ErrImagePull: rpc error: code = Unknown desc = Error response from daemon: pull access denied for kfp-operator-stub-provider, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```

alexgeorgousis:
### Current problem

The `provider` workflow step fails with:

```
Error (exit code 1): pods "experiment-ununpfd7vnnfqxsiim3-846t6-1731469356" is forbidden: User "system:serviceaccount:kfp-operator-system:default" cannot patch resource "pods" in API group "" in the namespace "kfp-operator-system"
```

Makefile
```diff
@@ -58,12 +58,7 @@ integration-test-up: ## Spin up a minikube cluster for integration tests
 	# Proxy K8s API
 	kubectl proxy --port=8080 & echo $$! > config/testing/pids
 
-integration-test: manifests generate helm-cmd yq ## Run integration tests
```
chr12c, line 61: don't remove the `##`because then it won't render in `make help`

alexgeorgousis: Ah didn't realise, I thought it was just a comment 😂

Reverted in fa98065c

alexgeorgousis, line 61: Should we run manifests and/or generate somewhere during minikube-up? We currently do both in `integration-test-up`, but I don't understand why they're needed.

chr12c: I think targets should only be run if they are relevant to whatever's happening. If something actually requires the helm templates and yaml to be in sync then they should both be run. I spot `unit-test` and `decoupled-test` also run `manifests` and `generate`, so hopefully they're both required. If not then it's worth thinking about at which point they should be run together.

config/testing/integration-test-values.yaml
```diff
@@ -1,12 +1,48 @@
...
+    enabled: true
+  webhookCertificates:
+    provider: cert-manager
   argo:
     serviceAccount:
-      name: default
-      create: false
+      name: kfp-operator-argo
+      create: true
     compilerImage: kfp-operator-stub-provider
```
alexgeorgousis, line 16: I don't get why we use the stub provider image as the compiler image. If we want to stub the compiler, why don't we use a compiler stub image? Just for simplicity?

Makefile
alexgeorgousis: I moved them to `minikube-helm-install-operator` in `minikube.mk` where I think they make more sense (when we actually install the operator which is what uses the files generated by those targets).

alexgeorgousis:
We are closing this PR and the related issue because:

- The solutions to the latest blocker [outlined above](https://github.com/sky-uk/kfp-operator/pull/477#issuecomment-2674988478) don't work or aren't easy to implement
- We think the provider controller logic is sufficiently tested in unit tests
- We don't need the changes this PR is implementing for the integration tests to work once we are [calling the provider service API instead of the CLI](https://github.com/sky-uk/kfp-operator/issues/376) from the argo workflows

Merging the 2 minikube targets (`integration-test-up`, `minikube-up`) is still a good idea, but we've spent more time on it already than we believe it's worth, so we've decided to put it aside for the moment until further notice.

PR #478, opened 2025-02-12
Restructure common module

chr12c:
Closes # .

## Tasks

- [ ] ...
- [ ] Documentation updated


chr12c:
Closing this PR because attempting to move because there are circular dependencies holding it back from a full on migration of argo/common as described in https://github.com/sky-uk/kfp-operator/issues/484.
A smarter approach will be to start with migrating individual components one by one to see if anything breaks, e.g. moving just the k8s client config: https://github.com/sky-uk/kfp-operator/pull/482

PR #482, opened 2025-02-13
Move K8sClientConfig from argo/common to common

chr12c:
Connected to #464, #484

Move K8sClientConfig from argo/common to common because it is not argo specific.

## Tasks

- [x] Move K8sClientConfig from argo/common to common 
- [x] Update module dependencies


controllers/webhook/status_updater.go
```diff
@@ -4,6 +4,7 @@ import (
 	"context"
 	pipelinesv1 "github.com/sky-uk/kfp-operator/apis/pipelines/v1alpha6"
 	"github.com/sky-uk/kfp-operator/argo/common"
+	k8sConfigCommon "github.com/sky-uk/kfp-operator/common"
```
grahamia, line 7: I would say this should be common and the argo common one be something elselike argocommon?
common -> k8sConfigCommon seems wrong. Or rootcommon but dont like that personally

chr12c: I decided not to alias the argo common import because it's used significantly more in the file, but I shall alias it with `argocommon` and make `k8sConfigCommon` the true `common` then.

chr12c: https://github.com/sky-uk/kfp-operator/pull/482/commits/6857f51d0891cda78dde16d24c09f46ffc13ad44

TobyPinfold (approved):

PR #483, opened 2025-02-14
Add multi-compiler functionality

### LINKED ISSUE #466: Support multiple compiler frameworks

Currently compiler support is set in the helm on the workflow from manager values or defaulted to `kfp-operator-argo-kfp-compiler:version`

e.g. in the compiled workflow we have the following:
```
      image: {{ include "kfp-operator.compilerImage" . }}

{{- define "kfp-operator.compilerImage" -}}
{{ if .Values.manager.argo.compilerImage -}}
{{ include "kfp-operator.trimmedContainerRegistry" . }}{{ .Values.manager.argo.compilerImage }}:{{ .Chart.AppVersion }}
{{- else -}}
{{ include "kfp-operator.trimmedContainerRegistry" . }}kfp-operator-argo-kfp-compiler:{{ .Chart.AppVersion }}
{{- end -}}
{{- end -}}
```

so therefore we only support one compiler framework per operator deployment, currently TFX compile is built as part of the KFP Operator build.

It would be good if we could use different frameworks specified as part of the pipeline resource and always fallback to the default configured operator configured compiler

One of two options:

Add new framework field to pipeline CRD which can be used as part of the naming convention for the compiler image

e.g. 
```
kind: Pipeline
spec:
  framework: tfx
  image: repo/kfp-operator-quickstart:0.6.0-25-g5e41f9d
  provider: vai
  tfxComponents: pipeline.create_components
```
This would then cause the workflow to use compiler image `kfp-operator-compiler-tfx:version` if not defined would use the standard.

Problem with above is no control in the operator for what images that could be used, could be anything.
Another possible solution would be to control a mapping from framework to image within the controller configuration/crd.

In the configuration you could have something like the following (helm example):
```
      frameworks:
        {{ pull extra config from values override if any required }}
        default: {{ include "kfp-operator.compilerImage" . }} # keep the default to fallback on if framework not specified in pipeline resource
```

This would allow during reconcile to lookup the framework in the config to ensure it is correct for the deployment and then pass it through the compiled workflow.

grahamia:
Closes #466 .

Changes to workflow to use framework from pipeline resource file

## Tasks

- [X] add framework to pipeline crd
- [X] change quickstart to be tfx framework
- [X] change workflows to use framework on pipeline if set
- [X] fallback to default if not set
- [X] update to latest yq image
- [X] Documentation updated


helm/kfp-operator/templates/workflows/common.yaml
```diff
@@ -19,8 +19,8 @@ spec:
     script:
       command:
       - ash
-      image: mikefarah/yq:4.27.5
-      source: yq e '.{{`{{inputs.parameters.field}}`}}' '/provider-config.json'
+      image: mikefarah/yq:4.45.1
+      source: yq e -r '.{{`{{inputs.parameters.field}}`}}' '/provider-config.json'
```
grahamia, line 23: FYI -r stops yq wrapping string values in output with quotes

helm/kfp-operator/templates/workflows/common.yaml
```diff
@@ -35,8 +35,24 @@ spec:
...
       command:
       - ash
-      image: mikefarah/yq:4.27.5
-      source: yq e '.image' /resource-definition.yaml
+      image: mikefarah/yq:4.45.1
+      source: yq e -r '.image' /resource-definition.yaml
+    metadata:
+      {{- .Values.manager.argo.metadata | toYaml | nindent 6 }}
+    activeDeadlineSeconds: {{ .Values.manager.argo.stepTimeoutSeconds.default }}
+  - name: select-framework-compiler-image
```
grahamia, line 43: This is the main new bit which calculates the correct image to use and falls back if not set

argo/kfp-compiler/Makefile
```diff
@@ -42,4 +42,9 @@ build:
 
 WHEEL_VERSION=$(shell poetry version | cut -d ' ' -f 2)
 DOCKER_BUILD_EXTRA_PARAMS=--build-arg WHEEL_VERSION=${WHEEL_VERSION}
+
 include ../../docker-targets.mk
+
+docker-push-all:
+	$(MAKE) docker-push IMG=kfp-operator-argo-kfp-compiler
```
grahamia, line 49: Think once we do the changes to add new framework compilers into project we rename/drop the default.

Makefile
```diff
@@ -178,7 +178,7 @@ $(NEWLINE)
 endef
 endif
 
-INDEXED_YAML := $(YQ) e '{([.metadata.name, .kind] | join("-")): .}'
+INDEXED_YAML := $(YQ) e --no-doc '{([.metadata.name, .kind] | join("-")): .}'
```
grahamia, line 181: The --no-doc stops yq from outputting --- between all resources

config/manager/workflows/compiled.yaml
```diff
@@ -47,7 +48,7 @@ spec:
     initContainers:
     - args:
       - /shared
-      image: kfp-operator-argo-kfp-compiler:latest
+      image: "{{inputs.parameters.compiler-image}}"
```
chr12c, line 51: Single quotes to be consistent in this file

apis/pipelines/v1alpha6/pipeline_types.go
```diff
@@ -18,10 +18,12 @@ type PipelineSpec struct {
 	TfxComponents string            `json:"tfxComponents" yaml:"tfxComponents"`
```
aidandunlop, line 18: appreciate this would be a breaking change but now we have a framework field, `TfxComponents` should really just be `Components`

grahamia: Yes, we should probably raise a v1alpha7 epic and add this to that as a breaking change

aidandunlop: can you raise an issue to do that plz

docs-gen/content/docs/reference/resources/pipeline.md
```diff
@@ -20,16 +20,19 @@ spec:
...
+| `spec.provider`      | Specify which provider to use in create/update workflows.                                                                                        |
+| `spec.image`         | Container image containing TFX component definitions.                                                                                            |
+| `spec.tfxComponents` | Fully qualified name of the Python function creating pipeline components.                                                                        |
+| `spec.env`           | List of named objects. These will be provided to the `tfxComponents` function as environment variables.                                          |
+| `spec.beamArgs`      | List of named objects. These will be provided as `beam_pipeline_args` when compiling the pipeline.                                               |
+| `spec.framework`     | Optional. Sets a specific compiler image for the pipeline. e.g. `framework: tfx` then image used would be `kfp-operator-**tfx**-compiler:latest` |
```
aidandunlop, line 35:
```suggestion
| `spec.framework`     | Optional. Sets a specific compiler image for the pipeline. For example, using `framework: tfx` would mean the pipeline would be compiled with the `kfp-operator-**tfx**-compiler:latest` image |
```

aidandunlop: also maybe document the default here?

aidandunlop, line 35: We should probably add a new section to the docs anway, such as "Supported Pipeline Frameworks"

apis/pipelines/v1alpha6/pipeline_types.go
grahamia: created issue https://github.com/sky-uk/kfp-operator/issues/489

docs-gen/content/docs/reference/resources/pipeline.md
grahamia: Maybe add that once we do the next card of adding another framework? I will add a task to the issue.
Done - https://github.com/sky-uk/kfp-operator/issues/488

config/manager/workflows/compiled.yaml
grahamia: Every "image" I can find has double quotes apart from `image: kfp-operator-controller:latest` so think double is standard here.
e.g.
image: "{{inputs.parameters.provider-image}}"

So was following convension

helm/kfp-operator/templates/workflows/common.yaml
```diff
@@ -35,8 +35,24 @@ spec:
...
+      - name: resource-definition
+        path: /resource-definition.yaml
+        raw:
+          data: |
+            {{`{{workflow.parameters.resource-definition}}`}}
+    script:
+      command:
+      - ash
+      image: mikefarah/yq:4.45.1
+      source: yq e -r '.framework | select(. != "" and . != null) as $framework | (("{{ include "kfp-operator.trimmedContainerRegistry" . }}kfp-operator-" + $framework + "-compiler:{{ .Chart.AppVersion }}") // "{{ include "kfp-operator.compilerImage" . }}")' /resource-definition.yaml
```
TobyPinfold, line 55: would be good to have same comment here as previously mentioned for kustomize config version

aidandunlop (dismissed):
agree with @TobyPinfold's comments but approving in advance

config/manager/workflows/common.yaml
```diff
@@ -34,10 +34,26 @@ spec:
...
+        raw:
+          data: |
+            {{workflow.parameters.resource-definition}}
+    script:
+      command:
+      - ash
+      image: mikefarah/yq:4.45.1
+      source: yq e -r '.framework | select(. != "" and . != null) as $framework | (("kfp-operator-" + $framework + "-compiler:latest") // "kfp-operator-argo-kfp-compiler:latest")' /resource-definition.yaml
+    metadata:
+      {}
```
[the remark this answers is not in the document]
grahamia: I have pushed a change to remove the line wrap but this is the same in all workflow templates so I will raise a separate issue to get all removed as it does look like it isnt needed.
Raised - https://github.com/sky-uk/kfp-operator/issues/492
@chr12c agrees.

helm/kfp-operator/templates/workflows/common.yaml
grahamia: Done in https://github.com/sky-uk/kfp-operator/pull/483/commits/d76b9db2d72763f6908b35b6560bb12f85298884

config/manager/workflows/common.yaml
```diff
@@ -34,10 +34,26 @@ spec:
...
+      - name: resource-definition
+        path: /resource-definition.yaml
+        raw:
+          data: |
+            {{workflow.parameters.resource-definition}}
+    script:
+      command:
+      - ash
+      image: mikefarah/yq:4.45.1
+      source: yq e -r '.framework | select(. != "" and . != null) as $framework | (("kfp-operator-" + $framework + "-compiler:latest") // "kfp-operator-argo-kfp-compiler:latest")' /resource-definition.yaml
```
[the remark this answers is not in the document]
grahamia: Done in https://github.com/sky-uk/kfp-operator/pull/483/commits/d76b9db2d72763f6908b35b6560bb12f85298884

chr12c (approved):

TobyPinfold (approved):

grahamia:
Reworked to solution with pre configured pipeline frameworks

controllers/pipelines/internal/workflowfactory/pipeline_workflow_factory.go
```diff
@@ -1,17 +1,23 @@
...
+	"github.com/sky-uk/kfp-operator/controllers/pipelines/internal/workflowconstants"
 )
 
-type PipelineDefinitionCreator struct {
+type PipelineParamsCreator struct {
 	Config config.KfpControllerConfigSpec
 }
 
-func (pdc PipelineDefinitionCreator) pipelineDefinition(
+const defaultFramework = "default"
```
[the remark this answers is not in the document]
peter-hazell: ah ok this is just the key in the map which we set to 'default', was a bit confusing at first

apis/config/v1alpha6/kfp_controller_config_types.go
```diff
@@ -27,6 +27,7 @@ type KfpControllerConfigSpec struct {
 	DefaultExperiment      string                `json:"defaultExperiment,omitempty"`
 	RunCompletionTTL       *metav1.Duration      `json:"runCompletionTTL,omitempty"`
 	RunCompletionFeed      ServiceConfiguration  `json:"runCompletionFeed,omitempty"`
+	Frameworks             map[string]string     `json:"frameworks,omitempty"`
```
alexgeorgousis, line 30: Just an idea: We can create type aliases for the framework name and image so instead of `map[string]string` (which can be anything, and we just have to know that it's actually name -> docker image), we can have `map[FrameworkName]FrameworkImage`.

alexgeorgousis: But if that makes things complicated downstream for some reason, nevermind.

peter-hazell, line 30: think we've agreed on this name now but it's not obvious from the naming what this actually is and the type doesn't give it away either

`FrameworkCompilerImages` ?? or something more clear

maybe it doesn't matter as it's in the docs 🤷

grahamia: I dont personally like type aliases, even when you do it is still just a string and can be set to anything. Images in other places are currently just a string and an image resource could be `my-image` or `test/my-image` or `some-repo:4567/test/my-image`

As with Petes suggestion can rename Frameworks to be something more meaningful

grahamia: Will rename to `PipelineFrameworkImages` as this is the name used elsewhere

grahamia: Done in https://github.com/sky-uk/kfp-operator/pull/483/commits/a91c6d573ead5e56a889b66f66e301b1bc742d38

alexgeorgousis (approved):

apis/config/v1alpha6/kfp_controller_config_types.go
```diff
@@ -19,14 +19,15 @@ type KfpControllerConfig struct {
...
-	RunCompletionFeed      ServiceConfiguration  `json:"runCompletionFeed,omitempty"`
+	DefaultProvider         string                `json:"defaultProvider,omitempty"`
+	DefaultProviderValues   DefaultProviderValues `json:"defaultProviderValues,omitempty"`
+	WorkflowTemplatePrefix  string                `json:"workflowTemplatePrefix,omitempty"`
+	WorkflowNamespace       string                `json:"workflowNamespace,omitempty"`
+	Multiversion            bool                  `json:"multiversion,omitempty"`
+	DefaultExperiment       string                `json:"defaultExperiment,omitempty"`
+	RunCompletionTTL        *metav1.Duration      `json:"runCompletionTTL,omitempty"`
+	RunCompletionFeed       ServiceConfiguration  `json:"runCompletionFeed,omitempty"`
+	PipelineFrameworkImages map[string]string     `json:"frameworks,omitempty"`
```
peter-hazell, line 30:
```suggestion
	PipelineFrameworkImages map[string]string     `json:"pipelineFrameworkImages,omitempty"`
```
?

grahamia: Correct, done this in https://github.com/sky-uk/kfp-operator/pull/483/commits/647a765412277e78a1c81aa3658d36c35832f020 as needed changing in multiple places for this

alexgeorgousis (approved):

peter-hazell (approved):

aidandunlop (approved):

PR #505, opened 2025-03-03
Update provider-api specification

### LINKED ISSUE #504: Provider Service API returns the same contents as provider-output.yaml in Provider CLI
The state handler requires contents of provider-output in order to function properly. Implementation of provider service API did not take this into account.

Make provider service api curl requests return the same contents as provider CLI

We need to change the api service output to meet the specification of the existing provider cli. Thereby the responses should contain the same structure of 

```json
{
 "id": <some-id>,
 "providerError": <some-error>,
}
```

TobyPinfold:
Closes #504 

Updates the provider api specification to include:
- adds id field to response body for delete requests
- adds providerError field to response body for any error scenarios
- updates website documented spec to correlate with the changes above

## Tasks

- [x] Tests updated
- [x] Documentation updated

### DIFF
--- a/provider-service/base/pkg/server/server.go
+++ b/provider-service/base/pkg/server/server.go
@@ -2,6 +2,7 @@ package server
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
@@ -29,7 +30,7 @@ func createHandler(hr resource.HttpHandledResource) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		body, err := io.ReadAll(r.Body)
 		if err != nil {
-			http.Error(w, "Failed to read request body", http.StatusInternalServerError)
+			writeErrorResponse(w, errors.New("failed to read request body"), http.StatusInternalServerError)
 			return
 		}
 		defer r.Body.Close()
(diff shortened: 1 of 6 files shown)


provider-service/base/pkg/server/server.go
```diff
@@ -158,3 +145,24 @@ func Start(ctx context.Context, cfg config.Server, provider resource.Provider) e
 
 	return nil
 }
+
+func writeErrorResponse(w http.ResponseWriter, providerError error, statusCode int) {
```
chr12c, line 149: I don't think this is correct. According to how `provider_app.go` does it, the `id` is always populated. This means `id` cannot have `omitempty`, and the server has to mimick how the CLI writes the id.

chr12c: actually ignore. I think I'm wrong about that.

TobyPinfold: the Id can't be populated for a create request that fails, so what is it using the id for on error scenarios. To me it makes no sense for it to be required at all.

TobyPinfold: implemented a fix for these scenarios in [5f1daec](https://github.com/sky-uk/kfp-operator/pull/505/commits/5f1daec3e4858762e49a2eadafe6810f7c992fb4)

If no providerId is returned ie any error scenario. The existing provider id should be retained on the resource.

TobyPinfold: added in [e14927f](https://github.com/sky-uk/kfp-operator/pull/505/commits/e14927f6638785cbf4baa27c1456d22039a0e41d) all errors bar a marshalling error / on a create operation will return the id

grahamia (approved):

PR #508, opened 2025-03-06
V1beta1 - framework object in pipeline resource

### LINKED ISSUE #490: v1beta1 - Pipeline CRD - Make framework an object
## Description
add a new framework object and remove framework specific attributes from the root of the resource.
```
metadata:
  name: penguin-pipeline
spec:
  provider: kfp
  image: kfp-quickstart:v1
  framework:
    type: tfx
    parameters:
       components: pipeline.create_components
       beamArgs:
       - name: arg1
          value: value1
```
where framework is object with `tfx: string` and `parameters: map[string]*apiextensionsv1.JSON`

Root attributes which are TFX framework specific are tfcComponents and beamArgs, these should be move to the generic parameters section.

### Convert To (v1alpha# -> v1beta1)
- Check remainder annotation
- if empty
  - Assume must be tfx framework
  - move value from tfxComponents into parameters map.
  - move any beamArgs NameValue array from root into parameters map (can be empty)
- else
  - populate framework attribute

### Convert From
If type=tfx
- take value from parameters.components and put into tfxComponents
- take value from parameters.beamArgs and put into beamArgs
for any other type
- framework should be stored in a remainder annotation for later use in conversion back to v1beta1

### Other changes
- Pass new structure into workflow.
- TFX compiler change to lookup components from new location.
- update docs

grahamia:
Closes #490 .

## Tasks

- [X] add framework field and add from and to transitions
- [X] Documentation updated


provider-service/vai/internal/provider/job_builder_test.go
```diff
@@ -4,12 +4,12 @@ package provider
 
 import (
 	"fmt"
+	testutil "github.com/sky-uk/kfp-operator/provider-service/base/pkg/testutil"
```
chr12c, line 7:
```suggestion
	"github.com/sky-uk/kfp-operator/provider-service/base/pkg/testutil"
```

controllers/pipelines/internal/workflowfactory/pipeline_workflow_factory_test.go
```diff
@@ -48,34 +51,37 @@ var _ = Describe("PipelineParamsCreator", func() {
...
-				Expect(compilerConfig.BeamArgs).To(Equal(expectedBeamArgs))
 			})
 
-			It("creates valid YAML", func() {
-				configYaml, err := yaml.Marshal(compilerConfig)
+			It("creates valid JSON", func() {
+				configYaml, err := json.Marshal(compilerConfig)
+				Expect(err).NotTo(HaveOccurred())
+
+				println(string(configYaml))
```
chr12c, line 62:
```suggestion
```

apis/pipelines/hub/test_generators.go
```diff
@@ -27,12 +28,25 @@ func RandomPipeline(provider string) *Pipeline {
 }
 
 func RandomPipelineSpec(provider string) PipelineSpec {
+	randomParameters := make(map[string]*apiextensionsv1.JSON)
+	randomParameters["components"] = &apiextensionsv1.JSON{Raw: []byte((fmt.Sprintf(`"%s"`, RandomString())))}
```
chr12c, line 32: is it guaranteed that `components` exists as a key here?

controllers/pipelines/internal/workflowfactory/pipeline_workflow_factory_test.go
```diff
@@ -48,34 +51,37 @@ var _ = Describe("PipelineParamsCreator", func() {
...
+				Expect(resultMap["name"]).To(Equal("pipelineNamespace/pipelineName"))
+				Expect(resultMap["image"]).To(Equal("pipelineImage"))
 
-				Expect(m["name"]).To(Equal("pipelineNamespace/pipelineName"))
-				Expect(m["image"]).To(Equal("pipelineImage"))
-				Expect(m["framework"]).To(Equal("pipelineFramework"))
-				Expect(m["tfxComponents"]).To(Equal("pipelineTfxComponents"))
-				env := m["env"].([]interface{})
-				Expect(env[0]).To(Equal(map[interface{}]interface{}{
+				framework := resultMap["framework"].(map[string]interface{})
```
chr12c, line 73: I think this can also panic?

apis/pipelines/hub/conversions.go
```diff
@@ -0,0 +1,54 @@
+package v1beta1
+
+import (
+	"encoding/json"
+	"github.com/sky-uk/kfp-operator/apis"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+)
+
+func AddComponentsToFrameworkParams(tfxComponents string, framework *PipelineFramework) {
+	marshal, _ := json.Marshal(tfxComponents)
```
chr12c, line 10: is this just adding quotes? Also error is ignored

apis/pipelines/hub/conversions.go
```diff
@@ -0,0 +1,54 @@
...
+	framework.Parameters["components"] = &apiextensionsv1.JSON{Raw: marshal}
```
chr12c, line 11: if `components` isn't a key then this panics

apis/pipelines/hub/conversions_test.go
```diff
@@ -0,0 +1,133 @@
...
+			Expect(components).To(Equal(result))
+		})
+
+		Specify("returns empty string if components do not exist", func() {
+			framework := NewPipelineFramework("tfx")
+
+			components, err := ComponentsFromFramework(&framework)
+
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(components).To(Equal(""))
```
chr12c, line 92:
```suggestion
			Expect(components).ToBe(Empty())
```

apis/pipelines/hub/conversions_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	var _ = Describe("ComponentsFromFramework", func() {
+		Specify("returns empty string for empty framework", func() {
+			framework := NewPipelineFramework("tfx")
+			framework.Parameters = nil
+
+			components, err := ComponentsFromFramework(&framework)
+
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(components).To(Equal(""))
```
chr12c, line 68:
```suggestion
			Expect(components).ToBe(Empty())
```

apis/pipelines/hub/pipeline_types.go
```diff
@@ -13,21 +14,36 @@ import (
...
-	Framework     string            `json:"framework,omitempty" yaml:"framework"`
+	Provider  string            `json:"provider" yaml:"provider"`
+	Image     string            `json:"image" yaml:"image"`
+	Env       []apis.NamedValue `json:"env,omitempty" yaml:"env"`
+	Framework PipelineFramework `json:"framework" yaml:"framework"`
+}
+
+type PipelineFramework struct {
+	Type       string                           `json:"type" yaml:"type"`
+	Parameters map[string]*apiextensionsv1.JSON `json:"parameters" yaml:"parameters"`
```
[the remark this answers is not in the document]
chr12c: also trying to understand why the values are pointers. You have to do 2 checks everytime a key is looked up to avoid a panic:
1. check key exists
2. check value is not nil

apis/pipelines/hub/conversions.go
grahamia: And turn into []byte. I did read the implementation of Marshal and on passing in a string it is impossible to return an error so rather than returning I thought just ignore here. Can return if you want

apis/pipelines/hub/conversions.go
grahamia: This is setting that key value 😕

apis/pipelines/hub/pipeline_types.go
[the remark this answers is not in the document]
grahamia: The parameter map is initialised in the construction we are using so wont be nil ever.

Good question, thanks for asking. This allows for the most flexibility possible in the parameter map, it allows for say
```
{
  "properties": {
    "key1": "value",
    "key2": null
  }
}
```
As this is a free property map full flexibility is given, we have the same pointer in the provider also.

apis/pipelines/hub/test_generators.go
grahamia: I guess this should be changed to generate a random map. We generated it with these so in the tfx specific tests we didnt have to recreate.

controllers/pipelines/internal/workflowfactory/pipeline_workflow_factory_test.go
grahamia: Guess could handle the fact of framework not being set to give better error message in the test

apis/pipelines/hub/conversions.go
TobyPinfold: will also be escaping special chars, so better to use `json.Marshal` as has been done

apis/pipelines/hub/conversions.go
chr12c: correcto

apis/pipelines/hub/conversions.go
grahamia: error handled in https://github.com/sky-uk/kfp-operator/pull/508/commits/ebbf7619d839bbbd45c39b4de20a4bb2c3d1bfdc

controllers/pipelines/internal/workflowfactory/pipeline_workflow_factory_test.go
grahamia: no panic https://github.com/sky-uk/kfp-operator/pull/508/commits/ebbf7619d839bbbd45c39b4de20a4bb2c3d1bfdc

apis/pipelines/hub/pipeline_types.go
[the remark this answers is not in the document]
TobyPinfold: Given this ends up in the compilers, there is value in allowing for 'null'  (JSON) / 'None' (Python) / Option (other langs). It should be something users might want

apis/pipelines/hub/conversions.go
```diff
@@ -0,0 +1,58 @@
...
+	if beamArgsExists {
+		var res []apis.NamedValue
+		if err := json.Unmarshal(beamArgs.Raw, &res); err != nil {
+			return nil, err
+		}
+
+		return res, nil
+	} else {
+		return []apis.NamedValue{}, nil
+	}
```
TobyPinfold, line 57:
```suggestion
	}
	
	return []apis.NamedValue{}, nil
```

apis/pipelines/hub/conversions.go
```diff
@@ -0,0 +1,58 @@
...
+	components, componentsExists := framework.Parameters["components"]
+	if componentsExists {
+		var res string
+		if err := json.Unmarshal(components.Raw, &res); err != nil {
+			return "", err
+		}
+		return res, nil
+	} else {
+		return "", nil
+	}
```
TobyPinfold, line 40:
```suggestion
	}
	return "", nil
```

TobyPinfold (approved):

grahamia:
OK to test

chr12c (approved):

PR #512, opened 2025-03-10
Extract stub compiler from stub provider cli

### LINKED ISSUE #511: Decouple stub provider cli from stub compiler
The stub provider cli is the same image as the stub compiler. The provider cli should be removed (separate issue) since it's no longer used (superseded by provider service), but the stub compiler is still required.

Move the stub compiler out of the existing stub provider CLI. It might make sense to have it live in the `compilers` directory.

chr12c:
Closes #511
- extract stub compiler from stub provider cli
- remove select-provider-image and select-provider-sa argo workflow template steps because these were only required for the provider cli
- update integration tests to reference the stub compiler
- remove provider cli
- remove unused python kfp_ext job unit tests
- add json tags to provider-service resource definitions (caused integration test to be flakey when some string fields are randomed to be a bunch of numbers, the json unmarshaller does not know how to convert it to a string without the json tags because it's relying on the yaml tags).


compilers/stub/Makefile
```diff
@@ -0,0 +1,7 @@
+include ../../docker-targets.mk
+include ../../help.mk
+
+IMG := kfp-operator-stub-compiler
+
+##@ Build
+build:
```
aidandunlop, line 7: might just want to add a comment here explaining why we need this empty target

aidandunlop (approved):

compilers/stub/Makefile
chr12c: https://github.com/sky-uk/kfp-operator/pull/512/commits/d2b590d1db724e2ae6e964686f7da12e1fb9e9fb

PR #514, opened 2025-03-10
Remove caret range specifier from package.json to fix package versions

chr12c:
Closes # .
Experimental draft PR to check if the CI's `make website` will bring in new dependencies

## Tasks

- [ ] ...
- [ ] Documentation updated


chr12c:
removing caret range specifier still brought in dependencies on `make website`

PR #523, opened 2025-03-11
Use consistent JSON format across all workflow parameters

### LINKED ISSUE #519: Consistent data format for resources submitted to workflows (JSON instead of YAML)
**Background**

Currently all resources submitted to workflows are in a YAML format (apart from the provider resource which is JSON).

There are problems serialising the framework object into YAML as part of this issue when using the `apiextensionsv1.JSON` type as part of the parameters: https://github.com/sky-uk/kfp-operator/issues/490. Using JSON across the board instead of YAML will solve this problem.

**Acceptance Criteria**

- [ ] All resources submitted to workflows should be JSON not YAML

TobyPinfold:
Closes #519 

Changes all resources passed into workflows to be in JSON format, instead of YAML. 

## Tasks

- [x] Update resources to use JSON format
- [x] Update all workflows to pass resource json as an artifact
- [x] Remove yaml -> json conversion within workflows

### DIFF
--- a/config/manager/workflows/simple.yaml
+++ b/config/manager/workflows/simple.yaml
@@ -23,22 +23,17 @@ spec:
         valueFrom:
           parameter: '{{steps.provider.outputs.parameters.provider-output}}'
     steps:
-    - - name: convert-resource-definition-yaml-to-json
-        templateRef:
-          name: kfp-operator-common-steps
-          template: yaml-to-json
-        arguments:
-          parameters:
-          - name: yaml
-            value: '{{workflow.parameters.resource-definition}}'
     - - name: provider
         templateRef:
           name: kfp-operator-common-steps
           template: create
         arguments:
           artifacts:
-          - from: '{{steps.convert-resource-definition-yaml-to-json.outputs.artifacts.json}}'
-            name: body
+          - name: body
+            raw:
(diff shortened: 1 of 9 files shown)


peter-hazell (approved):

config/manager/workflows/simple.yaml
```diff
@@ -23,22 +23,37 @@ spec:
...
+    - - name: select-provider-sa
+        templateRef:
+          name: kfp-operator-common-steps
+          template: select-provider-parameter
+        arguments:
+          parameters:
+          - name: field
+            value: serviceAccount
+          - name: provider-config
+            value: '{{workflow.parameters.provider-config}}'
```
chr12c, line 45: why is these steps being readded? I removed it in https://github.com/sky-uk/kfp-operator/pull/512 because this image is the provider cli image, and the service account is for the cli workflows.

chr12c, line 45: why are these steps being readded? I removed it in https://github.com/sky-uk/kfp-operator/pull/512 because this image is the provider cli image, and the service account is for the cli workflows.

chr12c (approved):

peter-hazell (approved):

PR #532, opened 2025-03-19
VAI Provider handle pipeline schema 2.1

### LINKED ISSUE #533: v1beta1 - support pipeline schema version 2.1 in VAI provider
## Description

currently in the VAI providers create pipeline code we have and "enricher" that pulls values like labels, the pipelineSpec and name from the pipeline definition. This currently only supports schema version 2.0.

## Tasks
We need to support both 2.0 and 2.1 schema.
Pipeline root should not be retrieved from the pipeline definition it should be created from the pipeline root storage configuration value combined with the namespaced pipeline name.

- [ ] add schema 2.1 enricher
     - name is in pipelineInfo.name
     - no such thing as labels in schema 2.1, add schemaVersion and sdkVersion
     - pipelineSpec is the whole object
- [ ] pass pipeline root storage from provider controller into provider as env var
- [ ] use to populate the pipeline job submitted to VAI

grahamia:
Closes #533.

## Tasks

- [ ] handle 2.1
- [ ] Documentation updated


provider-service/vai/internal/provider/job_enricher.go
```diff
@@ -14,12 +16,41 @@ type JobEnricher interface {
...
 
-type DefaultJobEnricher struct{}
+type DefaultJobEnricher struct {
+	schema2JobEnricher  JobEnricher
+	schema21JobEnricher JobEnricher
+}
+
+type Schema2JobEnricher struct{}
+
+type Schema21JobEnricher struct{}
```
peter-hazell, line 26: Just a though I wonder if rather than having multiple enrichers like this

We have a single enricher that just takes the fields that PipelineJob needs

And we create some sort of schema parser that can parse either version of the schema and returns those fields that the enricher needs

This would separate the 'enriching' of the pipeline job from the parsing of different schema versions ?

grahamia: Kind of agree, will look to separate into two responsibilities

grahamia: Refactored

PR #542, opened 2025-03-26
V1beta1 status rework

chr12c:
Closes # .

## Tasks

- [ ] ...
- [ ] Documentation updated


apis/conditions.go
alexgeorgousis, line 1: I'm not sure if this is a problem, so I'm gonna flag it just in case:

Since we've extracted the conditions available on our CRDs to the top level of `apis/`, they aren't versioned anymore. So, if we make a change to `SynchronizationState` it will immediately take effect on all CRD versions.

Since we don't currently do it this way, I'm assuming we've made the decision that changing the `SynchronizationState` should not require introducing a new CRD version, is that right?

TobyPinfold: This is deliberate, the SynchronizationState between all current versions is the same, duplicating this everywhere is wasteful. Though this is more related to the Conditions which were not initially common and now are between versions as it follows the same pattern.

alexgeorgousis: I just realised that `SynchronizationState` was already in the top level `apis` package, so nevermind, we're not actually deviating from the current pattern, we're just making our condition types common as well.

controllers/pipelines/runconfiguration_controller.go
```diff
@@ -78,11 +77,11 @@ func (r *RunConfigurationReconciler) Reconcile(ctx context.Context, req ctrl.Req
...
-		runConfiguration.Status.SynchronizationState = apis.Failed
+		runConfiguration.Status.SetSynchronizationState(apis.Failed, StateHandlerConstants.ProviderChangedError)
 
 		message := fmt.Sprintf(
 			`%s: %s`,
-			string(runConfiguration.Status.SynchronizationState),
+			string(apis.Failed),
 			StateHandlerConstants.ProviderChangedError,
 		)
 		r.EC.Recorder.Event(runConfiguration, EventTypes.Warning, EventReasons.SyncFailed, message)
```
alexgeorgousis, line 87: Line 84: Should we read the synchronization state (`apis.Failed`) from the conditions instead of hardcoding it again? Or is there a reason we didn't do that?

controllers/pipelines/state_handler.go
```diff
@@ -98,41 +118,47 @@ func (st *StateHandler[R]) onUnknown(
...
+	status := SetStatus{
+		Status: pipelineshub.Status{
+			Version: newVersion,
 		},
+	}
+
+	status.WithSyncStateCondition(apis.Creating, transitionTime, "")
+
+	return []Command{
+		status,
```
alexgeorgousis, line 161: Minor naming suggestion: At the beginning of `stateTransition`, we create a `SetStatus` and call the variable `setStatus`. Here, we call it `cmd` first and then `status`. Is there a reason we didn't settle on one of these everywhere?

controllers/pipelines/runconfiguration_controller.go
TobyPinfold: We set it above to be apis.Failed, extracting this from the list of conditions is wasted effort when we have just explicitly set this to be failed.

alexgeorgousis: Fair enough, had the same thought - just thought I'd point it out.

I was actually mentioning it because there's a 3rd time below that it's done, but I realised that one's not fixable.

alexgeorgousis (approved):

PR #547, opened 2025-03-26
Provider service metrics endpoint

aidandunlop:
Closes https://github.com/sky-uk/kfp-operator/issues/551

This PR exposes a metrics endpoint in the provider services, to allow users to monitor the basic state of the provider services. It's added to the `base` server, so that all provider services can utilise it. The metrics endpoint port is configurable via the KFP Operator helm values.

This PR also: 
- extends the KFPControllerConfig configmap in the Helm chart to allow labels to be set on the provider service pods, deployments, and services.
- updates the minikube scripts to make it easier to upgrade the local development environment while testing changes.

## Tasks

- [x] Expose metrics endpoint
- [x] Documentation updated


alexgeorgousis (approved):

alexgeorgousis (approved):

config/crd/bases/config.kubeflow.org_kfpcontrollerconfigs.yaml
```diff
@@ -444,6 +444,8 @@ spec:
                     additionalProperties:
                       type: string
                     type: object
+                  metricsPort:
+                    type: integer
```
chr12c, line 448: where is service port

TobyPinfold: line 7000 something, its still there just hidden by the pod spec which takes up a few thousand lines

DEVELOPMENT.md
```diff
@@ -117,7 +117,7 @@ make integration-test-down
 
 ## Run locally
 
-Running `make minikube-up NAME=provider` (set `provider` to the name of your chosen provider e.g. `NAME=vai`) will spin up a local K8s cluster with the operator deployed along with your chosen provider.
+Running `make minikube-up` will spin up a local K8s cluster with the operator deployed.
```
peter-hazell, line 120: Not a problem with this PR as such - but since you've made a change here I'll mention it

on line 83 we have the heading `## Running locally` and we also have this here `# Run locally`

We should combine these in some way

provider-service/base/pkg/server/metrics_server_decoupled_test.go
```diff
@@ -0,0 +1,71 @@
...
+	Expect(err).NotTo(HaveOccurred())
+
+	metricsAddr := listener.Addr()
+
+	return &MetricsTestContext{ctx: ctx, Addr: metricsAddr, shutdown: shutdown}
+}
+func TestMetricsServerDecoupledSuite(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "Metrics Server decoupled Suite")
+}
```
TobyPinfold, line 41:
```suggestion

func TestMetricsServerDecoupledSuite(t *testing.T) {
	RegisterFailHandler(Fail)
	RunSpecs(t, "Metrics Server decoupled Suite")
}
```

provider-service/base/pkg/server/server.go
```diff
@@ -144,10 +144,19 @@ func newHandler(resources []resource.HttpHandledResource) http.Handler {
 	return mux
 }
 
-func Start(ctx context.Context, cfg config.Server, provider resource.Provider) error {
-	addr := fmt.Sprintf("%s:%d", cfg.Host, cfg.Port)
+func Start(ctx context.Context, cfg config.Config, provider resource.Provider) error {
+	addr := fmt.Sprintf("%s:%d", cfg.Server.Host, cfg.Server.Port)
 	logger := common.LoggerFromContext(ctx)
 
+	onShutdown, err := InitialiseMetricsServer(ctx, cfg.Metrics)
```
TobyPinfold, line 151: I don't think this should be initialised here, it should be within main or there should be a new structure
```
 | server
    |- provider_server.go
    | # contains the functionality to spin up the provider http server
    |- metrics_server.go
    | # contains the functionalitu to spin up the metrics server
    |_ server.go
      # helper to spin start both servers and used within main.go

```

provider-service/base/pkg/server/metrics_server.go
```diff
@@ -0,0 +1,79 @@
...
+	return func() {
+		if err := meterProvider.Shutdown(ctx); err != nil {
+			logger.Error(err, "Error shutting down metrics server")
+		}
+	}, nil
+}
+
+func newResource() *resource.Resource {
+	return resource.NewWithAttributes(semconv.SchemaURL,
+		semconv.ServiceName("kfp-operator"),
```
peter-hazell, line 55:
```suggestion
		semconv.ServiceName("provider-service"),
```

since is the provider service we are going to be exposing metrics for, perhaps this naming is clearer

as I understand it, this name needs to match whichever name is used in the otel meter to expose custom metrics e.g. response codes, latency etc. which we'll need to add later on in the provider service

controllers/pipelines/provider_service_manager.go
```diff
@@ -77,18 +79,28 @@ func (sm ServiceManager) Get(ctx context.Context, owner *pipelinesv1.Provider) (
...
-	}}
+	matchLabels := map[string]string{AppLabel: prefixedProviderName}
+	labels := MapConcat(sm.config.DefaultProviderValues.Labels, matchLabels)
+	ports := []corev1.ServicePort{
+		{
+			Name:       "http",
+			Port:       int32(sm.config.DefaultProviderValues.ServicePort),
+			TargetPort: intstr.FromInt(sm.config.DefaultProviderValues.ServicePort),
+			Protocol:   corev1.ProtocolTCP,
+		},
```
peter-hazell, line 90: could create a private constructor function like so

```go
func tcpServicePort(name string, port int) corev1.ServicePort {
    return corev1.ServicePort{
        Name:       name,
        Port:       int32(port),
        TargetPort: intstr.FromInt(port),
        Protocol:   corev1.ProtocolTCP,
    }
}
```

and then in the list of ports just call

```go
ports := []corev1.ServicePort{
    tcpServicePort("http", sm.config.DefaultProviderValues.ServicePort),
    tcpServicePort("metrics", sm.config.DefaultProviderValues.MetricsPort),
}
```

which makes it a tad more readable/less duplication 🤷

minikube.mk
```diff
@@ -17,6 +17,10 @@ minikube-install-dependencies:
 minikube-helm-install-operator: helm-package-operator ./local/values.yaml
 	$(HELM) install -f ./local/values.yaml kfp-operator dist/kfp-operator-$(VERSION).tgz --set containerRegistry=localhost:5000/kfp-operator
 
+minikube-helm-upgrade-operator: helm-package-operator ./local/values.yaml
+	$(HELM) upgrade -f ./local/values.yaml kfp-operator dist/kfp-operator-$(VERSION).tgz --set containerRegistry=localhost:5000/kfp-operator
+
+
```
peter-hazell, line 23: stray new line

provider-service/base/pkg/server/metrics_server.go
TobyPinfold: I would suggest this is driven from the config provided by the specific implementation

provider-service/base/pkg/server/metrics_server_decoupled_test.go
aidandunlop: addressed in 74d759b8b05d497194673c4b1016b253d018694c

controllers/pipelines/provider_service_manager.go
```diff
@@ -76,23 +77,18 @@ func (sm ServiceManager) Get(ctx context.Context, owner *pipelineshub.Provider)
 
 }
 
+
+func tcpServicePort(name string, port int) corev1.ServicePort {
+	return corev1.ServicePort{Name: name, Port: int32(port), TargetPort: intstr.FromInt(port), Protocol: corev1.ProtocolTCP}
+}
```
alexgeorgousis, line 82: Move this helper function to the bottom of the file, otherwise perfect! 🎉

alexgeorgousis (approved):

PR #553, opened 2025-04-01
Make storage version on CRDs dynamic

alexgeorgousis:
Closes # .

## Tasks

- [ ] ...
- [ ] Documentation updated


helm/kfp-operator/values.yaml
```diff
@@ -12,6 +12,7 @@ namespace:
   metadata: {}
 
 manager:
+  storedVersion: v1alpha6
```
alexgeorgousis, line 15: We decided to set the `storedVersion` in the default `values.yaml` to the current stable version (in this case, `v1alpha6`) so that if a user installs the operator from `master` using the default `values.yaml` they always get the stable version installed.

alexgeorgousis:
I tested the changes by upgrading the operator with `v1beta` set as the stored version (`v1alpha6` was the current version) and:
- Both versions were set as stored in the CRD status (even though `v1alpha6` had `storage: false`)
    ```
	storedVersions:
	  - v1alpha6
	  - v1beta1
    ```
- Getting resources already present in the cluster returned them as `v1alpha6`

aidandunlop (approved):

DEVELOPMENT.md
```diff
@@ -136,3 +136,16 @@ Steps to create a new version of the pipeline CRDs:
 1) Make the required schema changes in `hub`. This will involve changing the conversion code in all older versions.
 1) Register the old schema (e.g. `v1alpha1`) in the controller runtime (see [here](main.go#L56)).
 1) Copy changes into helm version of the CRD to match that generated in `config/crd/bases`.
+
+Steps to install a new CRD version:
+
+1) Make sure `.Values.manager.storedVersion` in the helm vlaues is set to the currently stable version (i.e. the version currently installed in the cluster). This ensures that the new version will not become the stored version until all changes for that version have been released.
```
grahamia, line 142:
```suggestion
1) Make sure `.Values.manager.storedVersion` in the helm values is set to the currently stable version. This ensures that the new version will not become the stored version until all changes for that version have been released.
```

grahamia:
Looks good apart from think you need to add something into the install guide - https://sky-uk.github.io/kfp-operator/versions/v0.6.0/getting-started/installation/ to cover the new value

alexgeorgousis:
> Looks good apart from think you need to add something into the install guide - [sky-uk.github.io/kfp-operator/versions/v0.6.0/getting-started/installation](https://sky-uk.github.io/kfp-operator/versions/v0.6.0/getting-started/installation/) to cover the new value

Hmmm yeah this is a bit annoying... I set it to v1alpha6 for now, but we might need to think about this a bit more from the POV of a user installing the operator with helm.

alexgeorgousis:
> > Looks good apart from think you need to add something into the install guide - [sky-uk.github.io/kfp-operator/versions/v0.6.0/getting-started/installation](https://sky-uk.github.io/kfp-operator/versions/v0.6.0/getting-started/installation/) to cover the new value
> 
> Hmmm yeah this is a bit annoying... I set it to v1alpha6 for now, but we might need to think about this a bit more from the POV of a user installing the operator with helm.
> 
> My thoughts are:
> 
> * If someone is reading the installation guide, they should be able to use the values.yaml file as-is
> * That means we'll need to either:
>   
>   * Always keep it up-to-date with the currently stable version
>   * Add a placeholder like `<version>` and have the user go and check what version they want (bad experience)
> * Even if we keep it up to date, `storedVersion` seems a bit too low-level for the installation guide, so maybe we should rename it

Ok we thought about it and have come up with the following solution:

- If `multiversion` is disabled (or not set at all, which is what happens if a user follows our installation instructions) then the currently stable version (at the moment `v1alpha6`) will be set as the stored version
- If `multiversion` is enabled, then the version specified in `manager.multiversion.storedVersion` will be set as the stored version

I've updated the PR description to reflect this.

helm/kfp-operator/test/values.yaml
```diff
@@ -1,6 +1,7 @@
 fullnameOverride: kfp-operator
 containerRegistry: ""
 manager:
+  storedVersion: v1beta1
```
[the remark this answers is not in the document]
alexgeorgousis: This is outdated, but the only difference is `storedVersion` has been moved under `manager.multiversion`.

helm/kfp-operator/values.yaml
alexgeorgousis: This is now outdated because, since `manager.multiversion.enabled` is set to `false` in `helm/kfp-operator/values.yaml`, the `storedVersion` field (now moved under `manager.multiversion` won't have an effect, so we've removed it from the file entirely.

Should we keep it in even though `multiversion.enabled` is set to `false`? Just to show what's available?

alexgeorgousis:
> > Looks good apart from think you need to add something into the install guide - [sky-uk.github.io/kfp-operator/versions/v0.6.0/getting-started/installation](https://sky-uk.github.io/kfp-operator/versions/v0.6.0/getting-started/installation/) to cover the new value
> 
> Hmmm yeah this is a bit annoying... I set it to v1alpha6 for now, but we might need to think about this a bit more from the POV of a user installing the operator with helm.
> 
> My thoughts are:
> 
> * If someone is reading the installation guide, they should be able to use the values.yaml file as-is
> * That means we'll need to either:
>   
>   * Always keep it up-to-date with the currently stable version
>   * Add a placeholder like `<version>` and have the user go and check what version they want (bad experience)
> * Even if we keep it up to date, `storedVersion` seems a bit too low-level for the installation guide, so maybe we should rename it

Ok new problem that this causes... If a user installs the operator without enabling multiversion, `v1beta` will not be installed. This is a problem because the controller code will always try to get the `hub` version of all resources, which is `v1beta`. So we need a better solution.

alexgeorgousis:
> > > Looks good apart from think you need to add something into the install guide - [sky-uk.github.io/kfp-operator/versions/v0.6.0/getting-started/installation](https://sky-uk.github.io/kfp-operator/versions/v0.6.0/getting-started/installation/) to cover the new value
> > 
> > 
> > Hmmm yeah this is a bit annoying... I set it to v1alpha6 for now, but we might need to think about this a bit more from the POV of a user installing the operator with helm.
> > My thoughts are:
> > 
> > * If someone is reading the installation guide, they should be able to use the values.yaml file as-is
> > * That means we'll need to either:
> >   
> >   * Always keep it up-to-date with the currently stable version
> >   * Add a placeholder like `<version>` and have the user go and check what version they want (bad experience)
> > * Even if we keep it up to date, `storedVersion` seems a bit too low-level for the installation guide, so maybe we should rename it
> 
> Ok new problem that this causes... If a user installs the operator without enabling multiversion, `v1beta` will not be installed. This is a problem because the controller code will always try to get the `hub` version of all resources, which is `v1beta`. So we need a better solution.

One solution is: Let users install `v1beta1` by default. I'll update this comment with details.

DEVELOPMENT.md
```diff
@@ -136,3 +136,16 @@ Steps to create a new version of the pipeline CRDs:
 1) Make the required schema changes in `hub`. This will involve changing the conversion code in all older versions.
 1) Register the old schema (e.g. `v1alpha1`) in the controller runtime (see [here](main.go#L56)).
 1) Copy changes into helm version of the CRD to match that generated in `config/crd/bases`.
+
+Steps to install a new CRD version:
+
+1) Make sure `.Values.manager.storedVersion` in the helm values is set to the currently stable version. This ensures that the new version will not become the stored version until all changes for that version have been released.
```
paoloambrosio-skyuk, line 142:
```suggestion
1) Make sure `.Values.manager.storedVersion` in the Helm values is set to the currently stable version. This ensures that the new version will not become the stored version until all changes for that version have been released.
```

docs-gen/content/docs/getting-started/installation.md
```diff
@@ -59,6 +59,7 @@ Valid configuration options to override the [Default `values.yaml`]({{< ghblob "
 | `manager.monitoring.serviceMonitor.create`                | Create a ServiceMonitor for the [Prometheus Operator](https://github.com/prometheus-operator/prometheus-operator)                                                                                                   |
 | `manager.monitoring.serviceMonitor.endpointConfiguration` | Additional configuration to be used in the service monitor endpoint (path, port and scheme are provided)                                                                                                            |
 | `manager.multiversion.enabled`                            | Enable multiversion API. Should be used in production to allow version migration, disable for simplified installation                                                                                               |
+| `manager.multiversion.storedVersion`                      | Must be set if `manager.multiversion.enabled` is set to `true`. It specifies which version of the CRDs should be set as the stored version.                                                                         |
```
paoloambrosio-skyuk, line 62: What about defaulting it to the latest? This way it becomes just an override and it avoidd having to update test values all the time.

DEVELOPMENT.md
```diff
@@ -136,3 +136,16 @@ Steps to create a new version of the pipeline CRDs:
 1) Make the required schema changes in `hub`. This will involve changing the conversion code in all older versions.
 1) Register the old schema (e.g. `v1alpha1`) in the controller runtime (see [here](main.go#L56)).
 1) Copy changes into helm version of the CRD to match that generated in `config/crd/bases`.
+
+Steps to install a new CRD version:
```
paoloambrosio-skyuk, line 140: Shouldn't this be in the main documentation? It's for _operators_, not developers.

DEVELOPMENT.md
alexgeorgousis: Done in 7240714b

docs-gen/content/docs/getting-started/installation.md
alexgeorgousis: Done (including docs updates) in 9d8ae851

DEVELOPMENT.md
alexgeorgousis: Ok so I ended up overhauling the documentation commit: 2cbb46fa

I didn't think the CRD version upgrade steps fit anywhere in the existing website structure, so I decided to introduce an `Operator Manual` section and add it there. Let me know what you think.

chr12c (approved):

paoloambrosio-skyuk (approved):

PR #560, opened 2025-04-08
Move triggered and observed pipeline version into triggers and dependencies respectively

### LINKED ISSUE #536: v1beta1 - Tidy run configuration status block
## Description

Use RC dependency and trigger status for pipelines by moving `(observed|triggered)PipelineVersion` into `status.(dependencies|triggers).pipeline.version`. i.e `observedPipelineVersion: mypipeline_fd0e1279-0a1650` and `triggeredPipelineVersion: mypipeline_fd0e1279-0a1650` exist at the root of the status block. 

- `runConfiguration.status.triggeredPipelineVersion` should be moved to `runConfiguration.status.triggers.pipeline.version` 
- `runConfiguration.status.observedPipelineVersion` should be moved to `runConfiguration.status.dependencies.pipeline.version`.

chr12c:
Closes #536 
Move
- `RunConfiguration.Status.TriggeredPipelineVersion` to `RunConfiguration.Status.Triggers.Pipeline.Version` 
- `RunConfiguration.Status.ObservedPipelineVersion` to `RunConfiguration.Status.Dependencies.ObservedPipelineVersion`
- `Run.Status.ObservedPipelineVersion` to `Run.Status.Dependencies.Pipeline.Version`
## Tasks

- [x] Conversions updated
- [x] Documentation updated


apis/pipelines/hub/runconfiguration_types.go
```diff
@@ -40,8 +40,9 @@ type RunSpecTriggerStatus struct {
 }
 
 type TriggersStatus struct {
-	RunConfigurations map[string]TriggeredRunReference `json:"runConfigurations,omitempty"`
-	RunSpec           RunSpecTriggerStatus             `json:"runSpec,omitempty"`
+	RunConfigurations        map[string]TriggeredRunReference `json:"runConfigurations,omitempty"`
+	RunSpec                  RunSpecTriggerStatus             `json:"runSpec,omitempty"`
+	TriggeredPipelineVersion string                           `json:"triggeredPipelineVersion,omitempty"`
```
aidandunlop, line 45: The [associated issue](https://github.com/sky-uk/kfp-operator/issues/536) says that it should be moved into `status.triggers.pipeline.version` rather than `runConfiguration.status.triggers.triggeredPipelineVersion`?

aidandunlop: and same for `Dependencies.ObservedPipelineVersion`

TobyPinfold (changes requested):
As @aidandunlop mentioned, current changes do not match the spec of the issue

apis/pipelines/hub/run_types.go
```diff
@@ -140,9 +140,13 @@ type RunReference struct {
 	Artifacts  []common.Artifact `json:"artifacts,omitempty"`
 }
 
+type ObservedPipeline struct {
```
chr12c, line 143: I cannot think of a better name than this because `Pipeline` already exists as a struct in the same package.

TobyPinfold: Given the constraint, I think this makes sense even though it does break the normal conventions when marshalling to / from json.

TobyPinfold: alternatively, given it currently only wraps the Version and would be identical to the RunConfigurations `PipelineTriggerStatus` so it could be shared as `VersionWrapper` or something similar

TobyPinfold (approved):

aidandunlop (approved):

apis/pipelines/hub/run_types.go
chr12c: I think this can be revisited and it won't be a breaking change if we decide to go with a shared struct later on (or rename the struct again) because the field tags are what determines the schema.

PR #561, opened 2025-04-08
Handle missing Runs / RunConfigurations in operator webhook

### LINKED ISSUE #493: Eventing webhook: better handling of resource lookup failures

The current process of an event being passed through a provider to the operator webhook [relies on the presence of the run / run configuration within the cluster](https://github.com/sky-uk/kfp-operator/blob/36341427767d1ac403713dcb8c93e574be19be8c/controllers/webhook/event_processor.go#L89). 

This is problematic because if the run / run configuration has been deleted / removed or is otherwise inaccessible by a user of the operator, then the webhook will [return a 400 (Bad Request) error](https://github.com/sky-uk/kfp-operator/blob/36341427767d1ac403713dcb8c93e574be19be8c/controllers/webhook/runcompletion_feed.go#L100). This error is misleading, since there might not be anything wrong with the request. Also, this error results in a constant loop of retries from the provider's data source (e.g. Pub/Sub).

A specific error code should be returned from the webhook referring to the non-existence of a resource to allow the client (provider service) to handle appropriately, and separately to the other failure scenarios.

TobyPinfold:
Closes #493 

### Webhook returns 410 for missing resources

If the webhook (run completion feed http server) cannot find a resource (`Run` or `RunConfiguration`) for an event, it returns a `410` http code, indicating the resource is gone and not coming back (e.g. it's been deleted or was never in the cluster in the first place).

### Pub/Sub messages get ACKed on 410 failures

The current setup for how messages get handled in the Pub/Sub source is:
- Success: `ACK`
- Failure: `NACK`

This PR changes this setup to:
- Success: `ACK`
- Unrecoverable failure: `ACK`
- Recoverable failure: `NACK`

The only unrecoverable failure we're introducing is the `410` case so, for now, Pub/Sub messaged get `ACK`ed in the case of success or `410` and `NACK`ed otherwise. This means that Pub/Sub won't resend the message if a resource is missing, but it will in all other failure scenarios.

## Tasks

 - [ ] ~~Does the constant retry loop issue exist with the kfp provider which uses the workflow source too? And if yes, should we then implement the same behaviour there?~~
   This work is done on the shared webhook sink component so this applies to both providers
- [x] Deploy and test manually

### DIFF
--- a/controllers/webhook/event_processor_unit_test.go
+++ b/controllers/webhook/event_processor_unit_test.go
@@ -104,7 +104,7 @@ var _ = Context("ToRunCompletionEvent", func() {
 		})
 	})
 
-	When("neither run configuration nor run passed", func() {
+	When("neither run configuration or run passed", func() {
 		It("returns an error", func() {
 			runCompletionEventData := RandomRunCompletionEventData()
 			stubbedFilterFunc := func(_ []common.PipelineComponent, _ []pipelineshub.OutputArtifact) []common.Artifact {
@@ -112,7 +112,7 @@ var _ = Context("ToRunCompletionEvent", func() {
 			}
 
 			_, err := ResourceArtifactsEventProcessor{filter: stubbedFilterFunc}.ToRunCompletionEvent(&runCompletionEventData, nil, nil)
-			Expect(err).To(HaveOccurred())
+			Expect(err).To(Equal(&InvalidEvent{err.Error()}))
 		})
 	})
 })
--- a/controllers/webhook/event_processor_unit_test.go
+++ b/controllers/webhook/event_processor_unit_test.go
@@ -3,60 +3,28 @@
 package webhook
 
 import (
-	"context"
-
-	"github.com/go-logr/logr"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	"github.com/sky-uk/kfp-operator/apis"
 	pipelineshub "github.com/sky-uk/kfp-operator/apis/pipelines/hub"
 	"github.com/sky-uk/kfp-operator/argo/common"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
-func schemeWithCRDs() *runtime.Scheme {
-	scheme := runtime.NewScheme()
-
-	groupVersion := schema.GroupVersion{Group: "pipelines.kubeflow.org", Version: "v1beta1"}
-	scheme.AddKnownTypes(groupVersion, &pipelineshub.RunConfiguration{}, &pipelineshub.Run{})
-
-	metav1.AddToGroupVersion(scheme, groupVersion)
-	return scheme
-}
-
-func checkOutputArtifacts(outputArtifacts []pipelineshub.OutputArtifact, expectedArtifacts []pipelineshub.OutputArtifact) {
-	if outputArtifacts == nil {
-		Expect(expectedArtifacts).To(BeEmpty())
-	} else {
-		Expect(outputArtifacts).To(Equal(expectedArtifacts))
-	}
-}
-
 var _ = Context("ToRunCompletionEvent", func() {
-	var ctx = logr.NewContext(context.Background(), logr.Discard())
-
-	When("given valid runCompletionEventData", func() {
-		It("converts to a runCompletionEvent with filtered artifacts", func() {
+	When("run configuration passed and no run passed", func() {
+		It("returns RunCompletionEvent with filtered run configuration artifacts", func() {
 			rc := pipelineshub.RandomRunConfiguration(common.RandomNamespacedName())
+			rc.Spec.Run.Artifacts = []pipelineshub.OutputArtifact{{Name: "runconfig-artifact"}}
 
 			runCompletionEventData := RandomRunCompletionEventData()
 			runCompletionEventData.RunConfigurationName = &common.NamespacedName{
 				Name:      rc.Name,
 				Namespace: rc.Namespace,
 			}
+			runCompletionEventData.RunName = nil
 
-			expectedArtifacts := apis.RandomNonEmptyList(common.RandomArtifact)
+			expectedArtifacts := []common.Artifact{{Name: "runconfig-artifact"}}
 
-			stubbedFilterFunc := func(_ []common.PipelineComponent, _ []pipelineshub.OutputArtifact) []common.Artifact {
-				return expectedArtifacts
-			}
-
-			fakeClient := fake.NewClientBuilder().WithScheme(schemeWithCRDs()).WithObjects(rc).Build()
-
-			result, err := ResourceArtifactsEventProcessor{client: fakeClient, filter: stubbedFilterFunc}.ToRunCompletionEvent(ctx, runCompletionEventData)
+			result, err := ResourceArtifactsEventProcessor{filter: stubbedFilterFunc}.ToRunCompletionEvent(&runCompletionEventData, rc, nil)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(result).To(Equal(&common.RunCompletionEvent{
 				Status:                runCompletionEventData.Status,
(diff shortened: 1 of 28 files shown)


controllers/webhook/runcompletion_feed.go
```diff
@@ -64,52 +67,115 @@ func getRequestBody(ctx context.Context, request *http.Request) ([]byte, error)
...
+
+		for _, handler := range rcf.eventHandlers {
+			err := handler.Handle(*event)
+			if err != nil {
+				logger.Error(err, "Run completion event handler operation failed")
+				if k8sErrors.IsNotFound(err) {
+					http.Error(response, err.Error(), http.StatusGone)
+					return
+				} else {
 					http.Error(response, err.Error(), http.StatusInternalServerError)
```
alexgeorgousis, line 173: Here we return `http.StatusInternalServerError`, but none of the error type currently defined in `errors.go` returns such an error, so we probably need to add one to cover this case.

controllers/webhook/status_updater.go
```diff
@@ -156,7 +162,6 @@ func (su StatusUpdater) completeRunConfiguration(
 				"Action",
 				"Get",
 			)
-			return nil
```
alexgeorgousis, line 159: Hmmm I don't get why we've removed the `return nil` and instead let this return an error. I also don't get why we haven't done the same thing with the not found error below when updating the status.

alexgeorgousis:
I tested it manually and the webhook returns a `410` if the `RunConfiguration` doesn't exist at the time the VAI run finishes 🥳 (I just deleted the `RunConfiguration` manually while the run was running.

Next up: Does the Provider Service ACK the pubsub message when it receives a `410`?

controllers/webhook/status_updater.go
TobyPinfold: we were explicitly ignoring the error if it was `not found` so it would just progress, the logic has been moved to the layer above with the changes in this pr

TobyPinfold: see Handle function as above

controllers/webhook/event_processor_unit_test.go
```diff
@@ -150,72 +194,12 @@ var _ = Context("filter", func() {
 	})
 })
 
-var _ = Context("extractResourceArtifacts", func() {
```
TobyPinfold, line 153: related code is now deleted

controllers/webhook/event_processor_unit_test.go
```diff
@@ -71,6 +103,18 @@ var _ = Context("ToRunCompletionEvent", func() {
 
 		})
 	})
+
+	When("neither run configuration nor run passed", func() {
+		It("returns an error", func() {
```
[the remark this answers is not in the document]
TobyPinfold: fixed in [f1cd497](https://github.com/sky-uk/kfp-operator/pull/561/commits/f1cd497c5418c704010735df1146ccebcfdd4df8)

provider-service/base/pkg/streams/sinks/webhook_sink_unit_test.go
```diff
@@ -59,16 +64,40 @@ var _ = Context("SendEvents", func() {
...
 
-	When("webhook sink receives an invalid StreamMessage", func() {
-		It("should call its `OnFailureHandler` function", func() {
+	When("receives a recoverable error response from webhook server", func() {
+		It("should call the message's OnRecoverableFailureHandler", func() {
+			client := resty.New()
+			httpmock.ActivateNonDefault(client.GetClient())
+			webhookUrl := "/operator-webhook"
+			recoverableResponseCode := http.StatusInternalServerError
+			httpmock.RegisterResponder("POST", webhookUrl, httpmock.NewStringResponder(recoverableResponseCode, ""))
```
chr12c, line 73:
```suggestion
			httpmock.RegisterResponder(http.MethodPost, webhookUrl, httpmock.NewStringResponder(recoverableResponseCode, ""))
```

controllers/webhook/runcompletion_feed_decoupled_test.go
```diff
@@ -46,6 +53,16 @@ func (m MockRCEHandler) Handle(event common.RunCompletionEvent) error {
 	return nil
 }
 
+func schemeWithCRDs() *runtime.Scheme {
+	scheme := runtime.NewScheme()
+
+	groupVersion := schema.GroupVersion{Group: "pipelines.kubeflow.org", Version: "v1beta1"}
```
chr12c, line 59: `hub.GroupVersion`

controllers/webhook/status_updater.go
```diff
@@ -43,28 +42,54 @@ func NewStatusUpdater(ctx context.Context, scheme *runtime.Scheme) (StatusUpdate
...
 	logger.Info("StatusUpdater created")
-	return StatusUpdater{ctx, k8sClient}, nil
+	return StatusUpdater{k8sClient}, nil
 }
 
-func (su StatusUpdater) Handle(
-	event argocommon.RunCompletionEvent,
-) error {
+func (su StatusUpdater) Handle(ctx context.Context, event argocommon.RunCompletionEvent) EventError {
+	logger := log.FromContext(ctx)
```
chr12c, line 51: `log.FromContext(ctx).WithValues(...)` and put the run id and action into it. Remove them from below after

provider-service/base/pkg/streams/sinks/webhook_sink_unit_test.go
```diff
@@ -59,16 +64,40 @@ var _ = Context("SendEvents", func() {
...
+
+	When("receives an unrecoverable error response from the webhook server", func() {
+		It("should call the message's OnUnrecoverableFailureHandler", func() {
 			client := resty.New()
 			httpmock.ActivateNonDefault(client.GetClient())
 			webhookUrl := "/operator-webhook"
-			someNon200ResponseCode := 500
-			httpmock.RegisterResponder("POST", webhookUrl, httpmock.NewStringResponder(someNon200ResponseCode, ""))
+			unrecoverableResponseCode := http.StatusGone
+			httpmock.RegisterResponder("POST", webhookUrl, httpmock.NewStringResponder(unrecoverableResponseCode, ""))
```
chr12c, line 97:
```suggestion
			httpmock.RegisterResponder(http.MethodPost, webhookUrl, httpmock.NewStringResponder(unrecoverableResponseCode, ""))
```

TobyPinfold:
ok to test

controllers/webhook/runcompletion_feed.go
```diff
@@ -64,60 +65,121 @@ func getRequestBody(ctx context.Context, request *http.Request) ([]byte, error)
...
+
+func (rcf RunCompletionFeed) fetchRunConfiguration(ctx context.Context, name *common.NamespacedName) (*pipelineshub.RunConfiguration, EventError) {
+	logger := common.LoggerFromContext(ctx)
+	if name != nil {
+		rc := &pipelineshub.RunConfiguration{}
+		if err := rcf.client.Get(ctx, client.ObjectKey{
+			Namespace: name.Namespace,
+			Name:      name.Name,
+		}, rc); err != nil {
+			logger.Error(err, "failed to load", "RunConfig", name)
```
grahamia, line 90: should this be a warn rather than error

controllers/webhook/runcompletion_feed.go
```diff
@@ -64,60 +65,121 @@ func getRequestBody(ctx context.Context, request *http.Request) ([]byte, error)
...
-			http.Error(response, err.Error(), http.StatusBadRequest)
-			return
-		} else {
+		return run, nil
+	}
+
+	return nil, nil
+}
+
+func (rcf RunCompletionFeed) handleEvent(ctx context.Context) func(responseWriter http.ResponseWriter, request *http.Request) {
```
grahamia, line 122: this is rather a large function now... wondering if worth splitting up?

controllers/webhook/runcompletion_feed.go
TobyPinfold: there is only error or info, no warn without introducing verbosity levels to the mix

controllers/webhook/runcompletion_feed.go
TobyPinfold: issue is most of this code block is already calling functions and its just all the error handling that makes it so long. I would argue against splitting it up any further as it shows the full flow of logic for each case sequentially, which imo is easier to follow.

grahamia (approved):

TobyPinfold:
ok to test

TobyPinfold:
ok to test

PR #583, opened 2025-04-11
Remove expired start time from VAI schedules when it is in the past

### LINKED ISSUE #580: VAI triggering multiple runs when schedule created with a start date in the past
**Describe the bug**
Not really a bug, more of a unrequired feature of VAI.

**To Reproduce**
Submit a run configuration with a schedule trigger with a start date in the past. VAI will create as many runs as are missing to fill the schedule up, i.e. if you put a start date of 5 days in the past and the schedule was to run daily it would straight away create 5 runs.

```
spec:
  run:
    pipeline: pipeline
    provider: vai
  triggers:
    onChange:
    - pipeline
    schedules:
    - cronExpression: 0 11 * * *
      endTime: "T+5days"
      startTime: "T-5days"
```

**Expected behavior**
Just to start scheduling runs from now on, not backfill.

**Version and configuration**
VAI Provider

**Logs**
If applicable, add logs of the component to help explain your problem.

**Additional context**
One possible solution would be to check the start date on reconcile and fail if it is in the past.
Another would be to overwrite the startdate if it is in the past with now.

TobyPinfold:
Closes #580

There was behaviour reported that when submitting a schedule to Vertex AI with a `start_time` that has elapsed already, Vertex AI schedules all missed runs since the start_time. This changes ensures all schedules sent to vertex ai with a `start_time` set will only be in the future. If it is in the past, the field will be dropped from the request.

This has been tested and does prevent the behaviour reported from occurring, no "missed" runs are scheduled when a new schedule is deployed as we would normally expect.


provider-service/vai/internal/provider/job_builder.go
```diff
@@ -137,13 +138,14 @@ func (jb DefaultJobBuilder) MkSchedule(
 		AllowQueueing:         true,
 	}
 
-	if rsd.Schedule.StartTime != nil {
+	if rsd.Schedule.StartTime != nil && rsd.Schedule.StartTime.Time.After(time.Now()) {
```
chr12c, line 141: The reason of why this is done is worthy of adding to the doc comment

TobyPinfold: added in [cf50c3a](https://github.com/sky-uk/kfp-operator/pull/583/commits/cf50c3a7ceb739cc6b3079a03fe6dc0772052676)

chr12c (approved):

alexgeorgousis (approved):
Nice and simple 👌

TobyPinfold:
ok to test

PR #594, opened 2025-04-11
Decouple context from trigger server struct

### LINKED ISSUE #595: Decouple context from trigger server
According to [golang docs](https://pkg.go.dev/context) the context is not supposed to be stored in structs and reused per request. This could potentially affect metrics middlewares and the request itself.

Although no problems have already showed up, given that metrics are about to be implemented on the trigger service, it would be ideal to have a clean base with best practices on contexts to begin with.

Optionally also change the publisher to take a context per request, but since the context isn't being used in the publisher yet maybe defer to when metrics are implemented.
If/when we do any kind of tracing in the trigger service, the server would pass it's context to the publisher method call.

chr12c:
Closes #595

According to [golang docs](https://pkg.go.dev/context) the context is not supposed to be stored in structs and reused per request. Stop storing the context in trigger server and use the request context itself to initialise the per-request logger.

### DIFF
--- a/triggers/run-completion-event-trigger/internal/server/server.go
+++ b/triggers/run-completion-event-trigger/internal/server/server.go
@@ -15,13 +15,15 @@ import (
 
 type Server struct {
 	pb.UnimplementedRunCompletionEventTriggerServer
-	Context   context.Context
 	Config    *configLoader.Config
 	Publisher publisher.PublisherHandler
 }
 
-func (s *Server) ProcessEventFeed(_ context.Context, runCompletion *pb.RunCompletionEvent) (*emptypb.Empty, error) {
-	logger := common.LoggerFromContext(s.Context)
+func (s *Server) ProcessEventFeed(
+	ctx context.Context,
+	runCompletion *pb.RunCompletionEvent,
+) (*emptypb.Empty, error) {
+	logger := common.LoggerFromContext(ctx)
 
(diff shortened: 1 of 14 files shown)


triggers/run-completion-event-trigger/internal/server/server.go
```diff
@@ -15,37 +15,44 @@ import (
...
 		default:
-			return nil, status.Error(codes.Internal, "unexpected error occurred")
+			errMsg := "unexpected error occurred"
+			logger.Error(err, errMsg)
+			return nil, status.Error(codes.Internal, errMsg)
 		}
 	}
 
-	logger.Info("Run Completion Event Processed", "RunId", commonRunCompletionEvent.RunId)
+	logger.Info("Run Completion Event Processed")
```
alexgeorgousis, line 55: Why don't we wanna log the Run id anymore? Is it done elsewhere?

alexgeorgousis (approved):
Just left 1 minor comment - this is interesting... we probably do this (pass context to structs) in many places in the codebases so if this change goes well we can look to remove it elsewhere too

triggers/run-completion-event-trigger/internal/server/server.go
chr12c: It's baked into the logger the moment it's initialised, so there's no need to redeclare the run id every time it's used in this entire method.

PR #596, opened 2025-04-14
Remove `image` field from v1beta1 provider spec

### LINKED ISSUE #544: Remove `Image` from ProviderSpec
The `ProviderSpec.Image` isn't used anymore. It used to contain the image for the provider CLI, but now the operator has migrated over to using the provider service, this field is no longer used.

Remove the field from `ProviderSpec`.

TobyPinfold:
Closes #544.

Removed deprecated Image field from provider definition, and added conversion webhook between v1alpha6 and v1beta1 to manage the removal / addition of the field when transitioning versions.

## Tasks

- [x] Image field dropped from v1beta1 provider spec
- [x] Conversion webhook added between v1beta1 and v1alpha6
- [x] Tests updated
- [x] Documentation updated


apis/pipelines/hub/test_generators.go
```diff
@@ -63,25 +63,30 @@ func RandomPipelineSpec(provider common.NamespacedName) PipelineSpec {
...
+
 	return &Provider{
 		TypeMeta: metav1.TypeMeta{},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      RandomLowercaseString(),
 			Namespace: "default",
 		},
 		Spec:   RandomProviderSpec(),
-		Status: RandomStatus(common.NamespacedName{}),
+		Status: RemoveProvider(RandomStatus(common.NamespacedName{})),
```
chr12c, line 74: docstring to mention provider will never have a provider in the status.

chr12c:
Has this been manually verified

apis/pipelines/hub/test_generators.go
```diff
@@ -63,25 +63,32 @@ func RandomPipelineSpec(provider common.NamespacedName) PipelineSpec {
...
 	}
 }
 
+// RemoveProvider removes the provider from the status
+// This is because providers never have this field populated, nor should they.
```
aidandunlop, line 79: maybe just clarify that the status type is used by other resources hence why you can't just remove it from `RandomStatus`?
```suggestion
// This is because providers never have this field populated, nor should they, but share the Status type with other resources that do use this field.
```

alexgeorgousis (approved):

chr12c (approved):

apis/pipelines/hub/test_generators.go
chr12c: Shouldn't be sharing types when all the fields aren't being used in every type it's being shared with to begin with tbh

TobyPinfold: have added in latest changes

aidandunlop (approved):

PR #611, opened 2025-04-16
Providers to have allowed list of pipeline frameworks and handle default beam args

### LINKED ISSUE #531: v1beta1 - providers to have allowed list of pipeline frameworks and handle default beam args
## Description
As it currently stands, all providers would allow all pipeline compiler frameworks, even if the requested provider does not support the requested framework. We should allow the Provider to define what frameworks it supports.

We have also realised that the controller manager should not store any framework-specific config. We currently have the [images for the given framework set up in KfpControllerConfigSpec](https://github.com/sky-uk/kfp-operator/blob/master/apis/config/hub/kfp_controller_config_types.go#L30). These should be moved to the Provider CRD, so that the controller manager does not store anything framework specific.

Also, `defaultBeamArgs` is set at the top of the provider CRD, meaning it is global for all frameworks, despite only being used by the TFX compiler framework. This needs tidying up.

## Tasks
- [x] Change the provider CRD to include:
```yaml
frameworks:
- name: kfpsdk
  image: url to image
  patches:
  - type: json|merge
    patch: “some json patch string”
- name: tfx
  image: url to image
  patches:
  - type: json|merge
    patch: “some json patch string”
```
  - The patch types align with kubectl patch, the patch itself will always be a JSON payload
  - The JSON patches need to be applied by the pipeline reconciler to the framework parameters of the pipeline resource JSON before passing to the workflow

- Example of patch with before and after of pipeline frameworks JSON object:

**Patch JSON:**
```json
[
    {
      "op": "add",
      "path": "/Framework/parameters/beamArgs/-", # Adds to the end of the list
      "value": {
        "name": "newArg",
        "Value": "newValue"
      }
    }
]
```
**Before:**
```json
{
  "framework": {
    "type": "tfx",
    "parameters": {
      "components": "base_pipeline.create_components"
    }
  }
}
```
**After:**
```json
{
  "framework": {
    "type": "tfx",
    "parameters": {
      "components": "base_pipeline.create_components",
      "newArg": "newValue"
    }
  }
}
```

and drop DefaultBeamArgs 
- [x] Add conversion webhook for provider crd
- [x] Convert from v1alpha5/6 -> v1beta1 and back. Move default beam args to tfx supported framework.
- [x] Add validation to pipeline reconciler to ensure provider supports pipeline resources framework
- [x] In pipeline reconciler at call to trigger workflow need to apply the patches to the pipeline frameworks json
- [x] change compiler code to use new parameters
- [ ] update documentation

peter-hazell:
Closes #531

## Tasks

- [x] Added new `frameworks` field to provider CRD
- [x] Removed `defaultBeamArgs` from root of provider CRD
- [x] Updated TFX compiler to no longer use `defaultBeamArgs`
- [x] Added validation on pipeline reconciler to check that framework is supported by provider
- [x] Added JSON patching utils and patching of resource definitions as per defined provider patches
- [x] Documentation updated


apis/pipelines/v1alpha6/provider_conversion.go
aidandunlop, line 1: do we need to update v1alpha5's `provider_conversion.go` too?

peter-hazell: this is done now here: https://github.com/sky-uk/kfp-operator/pull/611/commits/e21144a5a1c54854307195d249987f9aa8cf5429

apis/pipelines/v1alpha5/provider_conversion.go
```diff
@@ -48,15 +87,30 @@ func (dst *Provider) ConvertFrom(srcRaw conversion.Hub) error {
...
+	if found {
+		beamArgs, err := hub.BeamArgsFromJsonPatches(tfxFramework.Patches)
+		if err != nil {
+			return err
+		}
+		dst.Spec.DefaultBeamArgs = beamArgs
+		dst.Spec.Image = tfxFramework.Image
+	} else {
+		return fmt.Errorf("tfx framework not in provider frameworks: %v", src.Spec.Frameworks)
+	}
```
TobyPinfold, line 103:
```suggestion
	if !found {
	        return fmt.Errorf("tfx framework not in provider frameworks: %v", src.Spec.Frameworks)
	}
	
	beamArgs, err := hub.BeamArgsFromJsonPatches(tfxFramework.Patches)
	if err != nil {
	    return err
        }
        
	dst.Spec.DefaultBeamArgs = beamArgs
	dst.Spec.Image = tfxFramework.Image
```

apis/pipelines/v1alpha6/provider_conversion.go
```diff
@@ -37,11 +77,25 @@ func (dst *Provider) ConvertFrom(srcRaw conversion.Hub) error {
 		return err
 	}
 
+	tfxFramework, found := common.Find(src.Spec.Frameworks, func(framework hub.Framework) bool {
+		return framework.Name == "tfx"
+	})
+
+	if found {
```
TobyPinfold, line 84: same pattern here

grahamia: Done here https://github.com/sky-uk/kfp-operator/pull/611/commits/f32d3632d5b99038289366397aa2dccc28c2087d

apis/pipelines/hub/provider_types.go
```diff
@@ -29,14 +29,25 @@ type ProviderList struct {
...
+type Patch struct {
+	Type  string `json:"type,omitempty" yaml:"type,omitempty"`
+	Patch string `json:"patch,omitempty" yaml:"patch,omitempty"`
+}
+
+type Framework struct {
+	Name    string  `json:"name,omitempty" yaml:"name,omitempty"`
+	Image   string  `json:"image,omitempty" yaml:"image,omitempty"`
+	Patches []Patch `json:"patches,omitempty" yaml:"patches,omitempty"`
+}
```
alexgeorgousis, line 41: For readability and to maintain consistency in the file, I think we should move the `Framework` and `Patch` types below the `ProviderSpec` type, so they're defined top-down: `ProviderSpec` > `Framework` > `Patch`.

apis/pipelines/hub/conversions.go
```diff
@@ -56,3 +57,36 @@ func BeamArgsFromFramework(framework *PipelineFramework) ([]apis.NamedValue, err
...
+func BeamArgsFromJsonPatches(patches []Patch) ([]apis.NamedValue, error) {
+	var namedValues []apis.NamedValue
+	for _, patch := range patches {
+		var patchOps []apis.JsonPatchOperation
+		err := json.Unmarshal([]byte(patch.Patch), &patchOps)
+		if err != nil {
+			return nil, err
+		}
+		for _, patchOp := range patchOps {
+			nameValue, ok := patchOp.Value.(map[string]interface{})
```
chr12c, line 70:
```suggestion
			nameValue, ok := patchOp.Value.(map[string]any)
```

aidandunlop (approved):

apis/pipelines/hub/conversions.go
```diff
@@ -56,3 +57,36 @@ func BeamArgsFromFramework(framework *PipelineFramework) ([]apis.NamedValue, err
...
+			if !ok {
+				return nil, fmt.Errorf("expected map[string]interface{}, got %v", patchOp.Value)
```
chr12c, line 72:
```suggestion
				return nil, fmt.Errorf("expected map[string]any, got %v", patchOp.Value)
```

apis/pipelines/hub/conversions.go
```diff
@@ -56,3 +57,36 @@ func BeamArgsFromFramework(framework *PipelineFramework) ([]apis.NamedValue, err
...
 }
+
+func BeamArgsFromJsonPatches(patches []Patch) ([]apis.NamedValue, error) {
+	var namedValues []apis.NamedValue
+	for _, patch := range patches {
+		var patchOps []apis.JsonPatchOperation
+		err := json.Unmarshal([]byte(patch.Patch), &patchOps)
+		if err != nil {
+			return nil, err
+		}
```
chr12c, line 68:
```suggestion
		if err := json.Unmarshal([]byte(patch.Patch), &patchOps); err ! = nil {
			return nil, err
		}
```

apis/pipelines/hub/conversions_test.go
```diff
@@ -132,4 +132,87 @@ var _ = Context("Conversions", func() {
...
+				{Type: "json", Patch: string(bytes)},
+			}
+
+			beamArgsFromPatches, err := BeamArgsFromJsonPatches(patches)
+
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(beamArgsFromPatches).To(Equal(beamArgs))
+		})
+
+		Specify("fail if a patch contains a value that isn't a map[string]interface{}", func() {
```
chr12c, line 174:
```suggestion
		Specify("fail if a patch contains a value that isn't a map[string]any", func() {
```

apis/pipelines/hub/provider_types.go
alexgeorgousis: Addressed in 5364f42a9304fd7cd0f2efb2b83a1396a6e75766, resolving comment ✅

apis/pipelines/hub/provider_types.go
```diff
@@ -34,9 +34,20 @@ type ProviderSpec struct {
...
+
+type Patch struct {
+	Type  string `json:"type,omitempty" yaml:"type,omitempty"`
+	Patch string `json:"patch,omitempty" yaml:"patch,omitempty"`
```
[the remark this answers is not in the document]
TobyPinfold: payload?

apis/pipelines/v1alpha5/provider_conversion.go
```diff
@@ -12,6 +15,42 @@ func (src *Provider) ConvertTo(dstRaw conversion.Hub) error {
...
+
+	tfxFramework := hub.Framework{
+		Name:  "tfx",
+		Image: DefaultTfxImage,
+	}
+
+	if len(beamArgsPatchOps) > 0 {
+		patchOpsBytes, err := json.Marshal(beamArgsPatchOps)
+		if err != nil {
+			return errors.New("failed to marshal patch operations to JSON")
```
chr12c, line 41: can you not just return the `err` here rather than create a new one. I presume the `err` already contains information about how it didn't marshal with potentially more info? That or append the contents of `err` in the new error message

TobyPinfold: for retained readability, but probs should use `fmt.Errorf("failed to marshal patch operations: %w", err)` to keep the error info

controllers/pipelines/internal/workflowfactory/pipeline_workflow_factory.go
```diff
@@ -12,16 +13,26 @@ import (
...
+type PipelineParamsCreator struct{}
+
+func findFramework(provider pipelineshub.Provider, pipeline *pipelineshub.Pipeline) (*pipelineshub.Framework, bool) {
+	requestedFramework := strings.ToLower(pipeline.Spec.Framework.Type)
 
-const defaultFramework = "default"
+	return apis.Find(provider.Spec.Frameworks, func(framework pipelineshub.Framework) bool {
+		return framework.Name == requestedFramework
+	})
+}
```
alexgeorgousis, line 24: Looking at the `findFramework` function I noticed we're comparing the framework `Type` field on the `Pipeline` to the framework `Name` field on the `Provider`.

Wouldn't it be better if the field had the same name, probably `Name`?

Otherwise, from the POV of a user creating a `Pipeline` resource, it's not obvious that the `type:` field on the `Pipeline` needs to match the `name:` field on the `Provider`.

Does this make sense?

controllers/pipelines/internal/jsonutil/json_utils_test.go
```diff
@@ -0,0 +1,129 @@
...
+			Expect(err).To(Not(HaveOccurred()))
+
+			patches := []pipelineshub.Patch{
+				{Type: JsonPatch, Patch: string(patchOpJson)},
+			}
+
+			result, err := PatchJson(patches, jsonBytes)
+			Expect(err).To(Not(HaveOccurred()))
+
+			var data map[string]interface{}
```
chr12c, line 40:
```suggestion
			var data map[string]any
```

controllers/pipelines/internal/jsonutil/json_utils_test.go
```diff
@@ -0,0 +1,129 @@
...
```
chr12c, line 65:
```suggestion
			var data map[string]any
```

controllers/pipelines/internal/jsonutil/json_utils_test.go
```diff
@@ -0,0 +1,129 @@
...
+			err = json.Unmarshal([]byte(result), &data)
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(data["bar"].([]interface{})[0].(string)).To(Equal("newValue"))
+			Expect(data["bar"].([]interface{})[1].(string)).To(Equal("b"))
```
chr12c, line 69:
```suggestion
			Expect(data["bar"].([]any)[0].(string)).To(Equal("newValue"))
			Expect(data["bar"].([]any)[1].(string)).To(Equal("b"))
```

apis/pipelines/v1alpha5/provider_conversion.go
```diff
@@ -12,6 +15,42 @@ func (src *Provider) ConvertTo(dstRaw conversion.Hub) error {
 	dst := dstRaw.(*hub.Provider)
 	dstApiVersion := dst.APIVersion
 
+	var beamArgsPatchOps []common.JsonPatchOperation
+
+	for _, nv := range src.Spec.DefaultBeamArgs {
```
TobyPinfold, line 20:
```suggestion
	for _, nameValue := range src.Spec.DefaultBeamArgs {
```

controllers/pipelines/internal/jsonutil/json_utils_test.go
```diff
@@ -0,0 +1,129 @@
...
+				{Type: JsonPatch, Patch: string(patchOpJson)},
+			}
+
+			result, err := PatchJson(patches, jsonBytes)
+			Expect(err).To(Not(HaveOccurred()))
+
+			var data map[string]interface{}
+			err = json.Unmarshal([]byte(result), &data)
+			Expect(err).To(Not(HaveOccurred()))
+			Expect(data["baz"].([]interface{})[0].(string)).To(Equal("hello"))
```
chr12c, line 94:
```suggestion
			var data map[string]any
			err = json.Unmarshal([]byte(result), &data)
			Expect(err).To(Not(HaveOccurred()))
			Expect(data["baz"].([]any)[0].(string)).To(Equal("hello"))
```

controllers/pipelines/internal/jsonutil/json_utils_test.go
```diff
@@ -0,0 +1,129 @@
...
+
+			patches := []pipelineshub.Patch{
+				{Type: MergePatch, Patch: mergeString},
+			}
+
+			result, err := PatchJson(patches, jsonBytes)
+			Expect(err).To(Not(HaveOccurred()))
+
+			println(result)
+			var data map[string]interface{}
```
chr12c, line 112:
```suggestion
			var data map[string]any
```

apis/pipelines/v1alpha5/provider_conversion.go
```diff
@@ -12,6 +15,42 @@ func (src *Provider) ConvertTo(dstRaw conversion.Hub) error {
...
+			Op:   "add",
+			Path: "/framework/parameters/beamArgs/0",
+			Value: map[string]string{
+				"name":  nv.Name,
+				"value": nv.Value,
+			},
+		}
+
+		beamArgsPatchOps = append(beamArgsPatchOps, patchOp)
+	}
```
TobyPinfold, line 31:
```suggestion
	for _, nameValue := range src.Spec.DefaultBeamArgs {
		patchOp := common.JsonPatchOperation{
			Op:   "add",
			Path: "/framework/parameters/beamArgs/0",
			Value: map[string]string{
				"name":  nameValue.Name,
				"value": nameValue.Value,
			},
		}

		beamArgsPatchOps = append(beamArgsPatchOps, patchOp)
	}
```

starting to think this nameValue could have a helper method on the stuct that just does a conversion to the map[string]string{} type

apis/pipelines/v1alpha6/provider_conversion.go
```diff
@@ -10,11 +14,47 @@ func (src *Provider) ConvertTo(dstRaw conversion.Hub) error {
...
+			Op:   "add",
+			Path: "/framework/parameters/beamArgs/0",
+			Value: map[string]string{
+				"name":  nv.Name,
+				"value": nv.Value,
+			},
+		}
+
+		beamArgsPatchOps = append(beamArgsPatchOps, patchOp)
+	}
```
TobyPinfold, line 30:
```suggestion
	for _, nameValue := range src.Spec.DefaultBeamArgs {
		patchOp := common.JsonPatchOperation{
			Op:   "add",
			Path: "/framework/parameters/beamArgs/0",
			Value: map[string]string{
				"name":  nameValue.Name,
				"value": nameValue.Value,
			},
		}

		beamArgsPatchOps = append(beamArgsPatchOps, patchOp)
	}
```

controllers/pipelines/internal/workflowfactory/experiment_workflow_factory.go
```diff
@@ -12,9 +12,10 @@ type ExperimentDefinitionCreator struct {
 }
 
 func (edc ExperimentDefinitionCreator) experimentDefinition(
+	_ pipelineshub.Provider,
```
chr12c, line 15: why is this unused parameter added here

chr12c: don't tell me it's to fulfil a type signature 😞

apis/utils.go
```diff
@@ -56,6 +56,15 @@ func Filter[T any](ts []T, filter func(T) bool) (filtered []T) {
 	return
 }
 
+func Find[T any](ts []T, predicate func(T) bool) (*T, bool) {
+	for _, v := range ts {
+		if predicate(v) {
+			return &v, true
+		}
+	}
+	return nil, false
```
TobyPinfold, line 65:
```suggestion
	for _, t := range ts {
		if predicate(t) {
			return &t, true
		}
	}
	return nil, false
```

pedantic but if its a range of ts then the value is a t 🤷‍♂️

controllers/pipelines/internal/workflowfactory/experiment_workflow_factory.go
chr12c: it is

controllers/pipelines/internal/jsonutil/json_utils.go
```diff
@@ -0,0 +1,43 @@
...
+import (
+	"fmt"
+	jsonpatch "github.com/evanphx/json-patch/v5"
+	pipelineshub "github.com/sky-uk/kfp-operator/apis/pipelines/hub"
+)
+
+const (
+	JsonPatch  = "json"
+	MergePatch = "merge"
+)
```
alexgeorgousis, line 12: If `JsonPatch` and `MergePatch` are the potential values of the `Type` field on the `Patch` type, then I think we should make that field an enum (it's currently a `string`) and define those enum variants in `provider_types.go` instead of this utils file.

apis/pipelines/hub/provider_types.go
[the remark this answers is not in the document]
grahamia: Will change to payload, dont really like document

apis/pipelines/v1alpha5/provider_conversion.go
grahamia: Agree with @TobyPinfold , will correct

apis/pipelines/v1alpha5/provider_conversion.go
grahamia: apart from this conversion code does it happen anywhere else?

apis/utils_test.go
```diff
@@ -221,3 +241,7 @@ var _ = Context("Utils", func() {
 		Entry("", map[string]int{"1": 1, "2": 1}, []int{1, 1}),
 	)
 })
+
+func ptr[T any](t T) *T {
```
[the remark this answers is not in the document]
grahamia: No.

apis/pipelines/v1alpha5/provider_conversion.go
TobyPinfold: doubtful....so yeah probs just easier to change and ignore the helper func

controllers/pipelines/internal/jsonutil/json_utils.go
TobyPinfold: If you mean use an iota to mimic an enum type, it gets messy because we are reading and exporting the jsonPatch / mergePatch as string fields we need to read.

The iota approach just turns it to an int, so we will end up converting to / from a string needlessly. 

The above makes sense if it were stored and retrieved in a db or something similar as theres no human having to discern 0 is jsonPatch and 1 is mergePatch. It's not as useful in our case where we are just reading whatever a user deploys.

controllers/pipelines/internal/workflowfactory/pipeline_workflow_factory.go
grahamia: I agree and will change to match. I noticed this myself when taking over the code and thought the same thing so glad someone else agrees

controllers/pipelines/internal/jsonutil/json_utils.go
alexgeorgousis: Ah I was actually thinking of the type alias approach (I believe we use this elsewhere already but not sure):

```golang
type Type string  // we can call it PatchType or something if Type is confusing (though it makes sense in context)

const (
    JsonPatch Type = "json"  // these could also just be called Json and Merge at that point potentially?
    MergePatch Type = "merge"
)
```

They'll still just be strings but what this gives us is:
1. The possible values of the type are defined next to the type (and not a random utils file)
2. We can use the type in func params etc instead of just `string`

grahamia: Not sure I see the point in a type alias for string? It still allows anything to be set in it `pipelineshub.Patch{Type: "some incorrect value", Payload: "patchie patch patch"}`  it is just a string after all?

grahamia: I agree it kind of makes sense to move these consts into where it is used, json_utils started out as being a standalone thing but is already linked to pipelineshub anyway

controllers/pipelines/internal/workflowfactory/pipeline_workflow_factory.go
```diff
@@ -12,16 +13,26 @@ import (
...
-	pipeline *pipelineshub.Pipeline,
-) (providers.PipelineDefinition, error) {
-	return providers.PipelineDefinition{
+	provider pipelineshub.Provider, pipeline *pipelineshub.Pipeline,
+) ([]pipelineshub.Patch, providers.PipelineDefinition, error) {
+	framework, found := findFramework(provider, pipeline)
+
+	if !found {
+		return nil, providers.PipelineDefinition{}, &workflowconstants.WorkflowParameterError{SubError: fmt.Sprintf("[%s] framework not support by provider", pipeline.Spec.Framework.Name)}
+	}
```
alexgeorgousis, line 33: I have a question about the approach we've taken to check if the framework specified on a `Pipeline` is supported by that `Pipeline`'s `Provider`:

We're currently checking this when we construct the (create or update) Argo Workflow, but it feels like a `Pipeline` reconciliation shouldn't get that far. It sounds to me like something that should be part of the resource validation logic, i.e. logic defined in the `pipeline_webhook.go` (currently the only resources with validation logic are `Runs` and `RunConfigurations`).

Is there a reason we decided not to do that (I actually don't have much experience with our validation webhooks so maybe there's something I don't know)?

grahamia: Providers could change so therefore it needs to be at workflow time unless we also add validation on the provider to stop frameworks from being changed if a pipeline exists that references a framework you are trying to remove?

docs-gen/content/docs/reference/resources/provider.md
```diff
@@ -12,14 +12,19 @@ Any referenced resources must always match the provider of the referencing resou
...
+| `spec.frameworks[0].name`               | Name of the framework.                                                                                                                                                                                                             | `tfx`                                                                                                               |
+| `spec.frameworks[0].image`              | Framework image.                                                                                                                                                                                                                   | `ghcr.io/kfp-operator/kfp-operator-tfx-compiler:version-tag`                                                        |
+| `spec.frameworks[0].patches`            | Patches to be applied to pipeline resource definition JSON.                                                                                                                                                                        |                                                                                                                     |
```
alexgeorgousis, line 25: I just wanted to give users a bit more detail about what these patches are and how they will be applied because I think this functionality is not that straightforward.

Sidenote: If you agree with the suggestion, you might want to apply it manually instead of committing directly from this comment so you can align the markdown table lines `|`.

```suggestion
| `spec.frameworks[0].patches`            | List of JSON patches that will be applied to every `Pipeline` resource that uses this `Provider`. The patches will be applied when the `Pipeline` resource is passed (as JSON) to the corresponding Argo Workflow.                                                                                                                                                                     |                                                                                                                     |
```

controllers/pipelines/internal/jsonutil/json_utils.go
chr12c: golang doesn't have good support for enum-like constructs. Even if you got the iota implementation, at some point you're going to do a case-switch against that enum with a default fall-back. Good support for enums means pattern matching would be exhaustive rather than requiring a fallback. You gain almost nothing by going for enums here.

TobyPinfold: We have this only once in the whole codebase, and using the type alias is arguable.

I do think we should have the kubebuilder enforcement though
```
// +kubebuilder:validation:Enum=json;merge
Type   string `json:"type,omitempty" yaml:"type,omitempty"`
```

docs-gen/content/docs/reference/resources/provider.md
```diff
@@ -12,14 +12,19 @@ Any referenced resources must always match the provider of the referencing resou
...
+| `spec.frameworks[0].patches[0].type`    | The type of patch to be applied to the pipeline resource definition JSON. Can be either `json` ([RFC6902](https://datatracker.ietf.org/doc/html/rfc6902)) or `merge` ([RFC7396](https://datatracker.ietf.org/doc/html/rfc7396)).   | `json`                                                                                                              |
+| `spec.frameworks[0].patches[0].payload` | The patch to be applied to the pipeline resource definition JSON.                                                                                                                                                                  | `[{ "op": "add", "path": "/framework/parameters/beamArgs/0", "value": { "name": "newArg", "value": "newValue" } }]` |
```
alexgeorgousis, line 27: Ok this might need a longer discussion because I'm now questioning the entire `patches` approach:

Every `Pipeline` has a framework with a `Name` and `Parameters`.

Couldn't we add a `Parameters` field to each `Framework` in the `Provider` which matches the `Parameters` field on `Pipelines`? Instead of generic JSON patches that can override anything on the `Pipeline` resource?

controllers/pipelines/internal/jsonutil/json_utils.go
alexgeorgousis: Nice, I didn't know about that kubebuilder validation 👍👍👍.

I still think the benefits I mentioned above are good enough to justify using a type alias:
> 1. The possible values of the type are defined next to the type (and not a random utils file)
> 2. We can use the type in func params etc instead of just string (which will be helpful to anyone reading the code, even if the compiler doesn't care)

But if you don't think those are good enough, fair - will leave the decision to you ✅

controllers/pipelines/internal/workflowfactory/pipeline_workflow_factory.go
alexgeorgousis: Hmmm yes I see...

Ok the way I'm thinking about it is like an API dependency: If the API (in this case a particular `Provider`) introduces a breaking change, then we should deal with that accordingly, right?

> Providers could change so therefore it needs to be at workflow time
When a framework in a `Provider` changes (in a breaking way), the `Pipelines` that depend on it will break, whether that happens at resource validation time or workflow creation time right? So what do we gain by moving the validation logic to workflow time (again, might be missing something about how validation webhooks work)?

grahamia: We will raise a new issue to cover adding webhook validation to pipeline and provider resources so that the underlying CRD changes can be releases now to unblock v1beta1.

docs-gen/content/docs/reference/resources/provider.md
grahamia: We discussed possible solutions in the design phase and agreed that using the patches solution give a much more generic solution allowing for all possible mutation of pipeline framework parameters by the provider operators. We can have defaults, overrides and anything else, e.g. remove a parameter that provider operators don't want users setting.

controllers/pipelines/internal/jsonutil/json_utils.go
grahamia: kubebuilder validation added.

controllers/pipelines/internal/workflowfactory/pipeline_workflow_factory.go
grahamia: Raised here - https://github.com/sky-uk/kfp-operator/issues/642

TobyPinfold (approved):

alexgeorgousis (approved):

PR #615, opened 2025-04-17
Include tfx framework image by default in manager config

alexgeorgousis:
## Overview

#508 introduced a breaking change for users who relied on the TFX framework image being specified in the manager `ConfigMap` as a `default` pipeline framework image.

This PR sets the TFX framework image as a default entry in `pipelineFrameworkImages` if no framework images are specified in `values.yaml`.

Note: This isn't an ideal solution since, if the KFP Operator is installed without the TFX framework image in `values.yaml`, it will not be included in `pipelineFrameworkImages` (so it's not a true default). However, this is only a temporary fix that will be undone in #531, where we'll move the supported pipeline framework config from the manager config to each `Provider` resource.


helm/kfp-operator/templates/_helpers.tpl
```diff
@@ -41,14 +41,6 @@ Remove trailing '/' characters from .Values.containerRegistry.
 {{- end }}
 {{- end }}
 
-{{- define "kfp-operator.compilerImage" -}}
-{{ if .Values.manager.argo.compilerImage -}}
-{{ include "kfp-operator.trimmedContainerRegistry" . }}{{ .Values.manager.argo.compilerImage }}:{{ .Chart.AppVersion }}
-{{- else -}}
-{{ include "kfp-operator.trimmedContainerRegistry" . }}kfp-operator-tfx-compiler:{{ .Chart.AppVersion }}
-{{- end -}}
-{{- end -}}
```
alexgeorgousis, line 50: This isn't used anywhere (it should have been removed in #508)

aidandunlop (approved):

peter-hazell (approved):

PR #619, opened 2025-04-24
Add new provider label to vai provider - including namespace and name

### LINKED ISSUE #618: Add provider label to submitted pipeline jobs/schedules

Currently within VAI or KFP there is no way to see which runs are from what provider, it would be useful to be able to do this.

On creating a schedule or one off run a label should be added by the provider with the provider name.

## Additional context

Additionally, this label could be used to filter run completion events within the VAI log router to split the events onto separate topics if required.

grahamia:
Closes #618 .

Adds extra labels when submitting resources to VAI, detailing which provider the resource was submitted from. This will be useful when dealing with multiple providers, as you will be able to filter logs and route logs using the provider label. 

## Tasks
- [x] add new provider-name label to submitted vai jobs

### DIFF
--- a/provider-service/vai/cmd/main.go
+++ b/provider-service/vai/cmd/main.go
@@ -57,7 +57,7 @@ func main() {
 }
 
 func runServer(ctx context.Context, vaiConfig *vaiConfig.VAIProviderConfig, baseConfig *baseConfig.Config) {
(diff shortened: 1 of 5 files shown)


chr12c (approved):

grahamia:
Force push was just a rebase of master but I somehow managed to lost @TobyPinfold last commit. These have been reapplied and retested. Reviews still stand
image

PR #624, opened 2025-04-25
Move all context fields out of structs in providers

### LINKED ISSUE #623: Move context out of provider structs

Currently lots of structs within the provider services are holding the context. This is not good practice and could cause issues if a context is stopped or changed during runtime. It is best practice to only store context in short lived structs and preferably not at all. 

See [go blog](https://go.dev/blog/context-and-structs) which gives further explanation on why.

We should extract the context out the structs and instead pass it as a parameter to the functions that require it. 

N/A

## Additional context

Any context, screenshots etc. about the feature request that you think is relevant.

TobyPinfold:
Closes #623 

## Tasks
- Removed all context fields within structs and moves them to be parameters passed through where necessary. [This addresses issues outlined in the official go documentation](https://go.dev/blog/context-and-structs)
- Use the http request's context as part of defining each http handler function

- [x] Documentation updated

### DIFF
--- a/provider-service/base/pkg/server/provider_server.go
+++ b/provider-service/base/pkg/server/provider_server.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/go-chi/chi/v5"
 	"github.com/go-chi/chi/v5/middleware"
+	"github.com/go-logr/logr"
 	"github.com/sky-uk/kfp-operator/argo/common"
 	"github.com/sky-uk/kfp-operator/provider-service/base/pkg/config"
 	"github.com/sky-uk/kfp-operator/provider-service/base/pkg/server/resource"
@@ -27,7 +28,11 @@ func livenessHandler(w http.ResponseWriter, _ *http.Request) {
 }
 
 func createHandler(ctx context.Context, hr resource.HttpHandledResource) http.HandlerFunc {
+	logger := common.LoggerFromContext(ctx)
+
 	return func(w http.ResponseWriter, r *http.Request) {
+		requestCtx := logr.NewContext(r.Context(), logger)
+
 		body, err := io.ReadAll(r.Body)
 		if err != nil {
 			writeErrorResponse(w, "", errors.New("failed to read request body"), http.StatusInternalServerError)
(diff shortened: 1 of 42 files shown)


provider-service/vai/internal/mocks/file_handler.go
```diff
@@ -1,21 +1,24 @@
...
+import (
+	"context"
+	"github.com/stretchr/testify/mock"
+)
 
 type MockFileHandler struct{ mock.Mock }
 
-func (m *MockFileHandler) Write(p []byte, bucket string, filePath string) error {
-	args := m.Called(p, bucket, filePath)
+func (m *MockFileHandler) Write(ctx context.Context, p []byte, bucket string, filePath string) error {
```
chr12c, line 10: I think for testify mocks you should be able to get away with ignoring the ctx by using `_` and removing it from `Called`

provider-service/kfp/internal/mocks/experiment_service.go
```diff
@@ -10,21 +11,23 @@ type MockExperimentService struct {
 }
 
 func (m *MockExperimentService) CreateExperiment(
+	ctx context.Context,
```
chr12c, line 14: same comment [here](https://github.com/sky-uk/kfp-operator/pull/624/files#r2063299640) applies to everything in `kfp/internal/mocks`

provider-service/vai/internal/mocks/file_handler.go
chr12c: https://github.com/sky-uk/kfp-operator/pull/624/commits/ad377c767d7ae0b322d44429d986d0932deb93d0

chr12c (approved):

chr12c:
OK to test

chr12c (approved):

provider-service/base/pkg/server/provider_server.go
```diff
@@ -26,16 +27,20 @@ func livenessHandler(w http.ResponseWriter, _ *http.Request) {
 	w.Write([]byte("Application is live."))
 }
 
-func createHandler(hr resource.HttpHandledResource) http.HandlerFunc {
+func createHandler(ctx context.Context, hr resource.HttpHandledResource) http.HandlerFunc {
+	logger := common.LoggerFromContext(ctx)
+
 	return func(w http.ResponseWriter, r *http.Request) {
+		requestCtx := logr.NewContext(r.Context(), logger)
```
chr12c, line 34: This is a significant enough of a change to be highlighted in the PR description. I think most people wouldn't have noticed the request itself held it's own context.

peter-hazell (approved):

PR #643, opened 2025-05-06
Compute pipeline hash across versions consistently

### LINKED ISSUE #530: Ignore reconcile on upgrade of api version of a resource.
## Is your feature request related to a problem? 

Currently when upgrading a resources api version, we immediately trigger reconciliation of the resource from its controller. This means any behaviour scheduled to apply on a `change` is triggered (i.e Pipeline change triggering a Run). For users with numerous pipelines this could be problematic as all pipelines will trigger their own run all at the same time of upgrade.

There needs to be a way to upgrade the resources to the latest version, whilst not automatically triggering reconciliations for the resources. This will help prevent triggering Runs when it is not required.

Potentially nice solution is:

1. On upgrading a resource using the webhook, set an annotation (i.e controller/upgraded-resource-via-webhook)
2. Create a new `WithEventFilter` predicate:
   - That identifies resources with the annotation and filters out its reconciliation requests so no reconcile is triggered. 
   - Remove the annotation from the resource
   - Equally ignore the operation of removing the annotation from the resource.

N/A

TobyPinfold:
Closes #530

Compute the pipeline hash across versions consistently. This ensures that for fields we have changed / added between spoke and hub versions, we ignore any default values we place in the hub, and ensure the hash is calculated in the same way as the spokes.

This ensures that the hash value will remain the same across version upgrades where no tangible change has actually been made to the pipeline spec, in turn preventing runs being triggered.

## Tasks

- [x] Update tests
- [x] Documentation updated


apis/test/cross_version_pipeline_hash_test.go
```diff
@@ -0,0 +1,92 @@
+package test
```
chr12c, line 1: test build tag

apis/pipelines/hub/pipeline_types.go
```diff
@@ -34,17 +36,63 @@ func NewPipelineFramework(compilerType string) PipelineFramework {
...
+		tfxComponentsList := []string{}
+		for _, v := range tfxComponents {
+			tfxComponentsList = append(tfxComponentsList, v)
+		}
 
-	output := make(map[string]string)
-	for key, value := range ps.Spec.Framework.Parameters {
-		raw, _ := json.Marshal(value.Raw)
-		output[key] = string(raw)
+		tfxComponentString = strings.Join(tfxComponentsList, "")
```
peter-hazell, line 57: join with empty separator?

apis/pipelines/hub/pipeline_types.go
```diff
@@ -34,17 +36,56 @@ func NewPipelineFramework(compilerType string) PipelineFramework {
 }
 
 func (ps Pipeline) ComputeHash() []byte {
```
chr12c, line 38: Add godoc to explain why here

apis/pipelines/hub/pipeline_types.go
TobyPinfold: [75b5120](https://github.com/sky-uk/kfp-operator/pull/643/commits/75b5120b1c62e95c1606a041041e6b85958d8973)

apis/test/cross_version_pipeline_hash_test.go
```diff
@@ -0,0 +1,137 @@
...
+			beta1PipelineVersion := beta1Pipeline.ComputeVersion()
+			alpha5pipelineVersion := alpha5Pipeline.ComputeVersion()
+			alpha6PipelineVersion := alpha6Pipeline.ComputeVersion()
+
+			Expect(alpha5pipelineVersion).To(Equal(beta1PipelineVersion))
+			Expect(alpha6PipelineVersion).To(Equal(beta1PipelineVersion))
+		})
+
+		Specify("Should compute the same version for identical pipelines when beamArgs aren't set in different api specs", func() {
+			delete(beta1Pipeline.Spec.Framework.Parameters, "beamArgs")
```
chr12c, line 109: i get what these deletes are for the purposes of these tests, but I don't agree that this then makes the order of these tests matter. Unless you specify the ordering.

TobyPinfold: Ordering is not related to the deletes, its the order of the hash writes.

TobyPinfold: Writing to the hash in a different order than the spokes, results in a different hash entirely.

TobyPinfold: solved, it uses a before each to reinitialise the pipelines for each test

chr12c (approved):

apis/pipelines/hub/pipeline_types.go
```diff
@@ -26,25 +26,76 @@ type PipelineFramework struct {
...
 		Name:       compilerType,
 		Parameters: make(map[string]*apiextensionsv1.JSON),
 	}
 }
 
+/*
+ComputeHash returns a hash of the pipeline spec, ensuring consistency across versions.
+
+To maintain compatibility, certain fields are deliberately excluded from the hash calculation, or are normalized to match
+the behavior of previous versions when using default fields as determined by conversion to the hub version.
```
peter-hazell, line 42:
```suggestion
the behaviour of previous versions when using default fields as determined by conversion to the hub version.
```

TobyPinfold: changed in [a87521b](https://github.com/sky-uk/kfp-operator/pull/643/commits/a87521b1f4ec9d314c91b957edd4800ccce0a6dd)

apis/pipelines/hub/pipeline_types.go
```diff
@@ -26,25 +26,76 @@ type PipelineFramework struct {
...
-	output := make(map[string]string)
-	for key, value := range ps.Spec.Framework.Parameters {
-		raw, _ := json.Marshal(value.Raw)
-		output[key] = string(raw)
+	// For the default framework, include components in the hash using the same format as spoke versions to ensure consistency.
+	tfxComponentsString := ""
+	if isFallbackFramework && pipeline.Spec.Framework.Parameters["components"] != nil {
+		if err := json.Unmarshal(pipeline.Spec.Framework.Parameters["components"].Raw, &tfxComponentsString); err != nil {
+			return nil
+		}
```
codehakase, line 61: nit: Rather than discard the error from unmarshalling the components meta, maybe add a debug log (note similar instances below too).

peter-hazell (approved):

apis/pipelines/hub/pipeline_types.go
TobyPinfold: I had considered this, exposing a log to identify there is an issue would be beneficial. There are drawbacks though to consider:

There currently is no logger exposed within this struct, and the interface it is adhering to for this computeVersion() function does not permit a logger or context to be passed through as a parameter.

Updating this will mean updating  the interface, in turn all implementations and many of the call sites will need reworking to expose the context also. It should be enough in this case to rely on the hash not being calculated to identify an issue for now 😃

PR #647, opened 2025-05-09
Configure allowed namespaces on Provider resources

### LINKED ISSUE #405: Configure allowed namespaces on Provider resources
## Overview

Currently, any resource can be deployed to any provider. The operator simply reads the specified provider on the resource spec and performs any provider operation it needs to perform (e.g. creating a training pipeline or schedule).

This raises security concerns, because it means any user can create and manage resources on any provider, without any checks to confirm if that user is allowed to use the specified provider.

We want to control which resources can be deployed to which providers, similarly to how `NetworkPolicies` are used to control which namespaces a service can accept requests from.

We also want to be aware of when resources attempt to access a provider that they are not allowed to use, to make incorrectly-configured resources/providers visible.

## Acceptance Criteria

- [x] Argo Workflows that perform provider CRUD operations can only be submitted for resources in namespaces permitted to use the specified provider
- [x] If a resource attempts to use a provider it's not permitted to use, the resource should go into a failed state and the user should receive an informative message
- [x] ~Attempts to use a provider that a resource is not permitted to use should increment a metric which can be used to detect if providers are being used maliciously or incorrectly configured~ This will be done as part of wider monitoring/metrics work

## Technical Details

Every time a request changes (created, updated, deleted) the following happens:
1. The operator receives a reconciliation request
2. The operator submits an Argo Workflow to the namespace the provider specified on the resource spec lives
3. The necessary operations are carried out in the specified provider (e.g. creating a training pipeline or a schedule)

Our goal is to only perform step 2 if the resource is in a namespace which is allowed to use the provider specified on the resource spec.

To do this we will add a list of allowed namespaces on `Provider` resources, which the operator will check against before submitting a Workflow to the specified provider. Something like the following (example taken from the [KFP Operator website](https://sky-uk.github.io/kfp-operator/docs/reference/resources/provider/)):

```yaml
kind: Provider
metadata:
  name: kfp
  namespace: kfp-operator
spec:
  image: kfp-operator-kfp-provider:<version>
  ...
  allowedNamespaces:
  - user-ns-1
  - user-ns-2
  - user-ns-3
```

If resource not in allowed list of namespaces then the status should be updated to failed with reason message.
If no allowedNamespaces list configured then no validation carried out.

peter-hazell:
Closes #405

This allows specifying the allowed namespaces in the provider resource which can be used at the point we create, update or delete workflows by validating the namespace of the resource against the allowed namespaces.

## Tasks

- [x] Updated provider CRD to include allowed namespaces
- [x] Added check at the point of workflow create, update and delete to verify that resource namespace is in the provider allowed namespaces


controllers/pipelines/internal/workflowfactory/resource_workflow_factory.go
```diff
@@ -127,6 +140,11 @@ func (workflows *ResourceWorkflowFactory[R, ResourceDefinition]) ConstructCreati
 		return nil, err
 	}
 
+	err = checkResourceNamespaceAllowed(resource.GetNamespacedName().Namespace, provider)
+	if err != nil {
```
grahamia, line 144:
```suggestion
	If err = checkResourceNamespaceAllowed(resource.GetNamespacedName().Namespace, provider); err != nil {
```

controllers/pipelines/internal/workflowfactory/resource_workflow_factory.go
```diff
@@ -193,6 +211,12 @@ func (workflows *ResourceWorkflowFactory[R, ResourceDefinition]) ConstructUpdate
 	if err != nil {
 		return nil, err
 	}
+
+	err = checkResourceNamespaceAllowed(resource.GetNamespacedName().Namespace, provider)
+	if err != nil {
```
grahamia, line 216:
```suggestion
	If err = checkResourceNamespaceAllowed(resource.GetNamespacedName().Namespace, provider); err != nil {
```

controllers/pipelines/internal/workflowfactory/resource_workflow_factory.go
```diff
@@ -260,6 +284,11 @@ func (workflows *ResourceWorkflowFactory[R, ResourceDefinition]) ConstructDeleti
 		return nil, err
 	}
 
+	err = checkResourceNamespaceAllowed(resource.GetNamespacedName().Namespace, provider)
+	if err != nil {
```
grahamia, line 288:
```suggestion
	If err = checkResourceNamespaceAllowed(resource.GetNamespacedName().Namespace, provider); err != nil {
```

peter-hazell: address here and also added resource name into the logging https://github.com/sky-uk/kfp-operator/pull/647/commits/c166d896cea9daa305bf13e1ab802a751ce86fcd

controllers/pipelines/internal/workflowfactory/resource_workflow_factory.go
peter-hazell: addressed here and also added resource name into the logging https://github.com/sky-uk/kfp-operator/pull/647/commits/c166d896cea9daa305bf13e1ab802a751ce86fcd

controllers/pipelines/internal/workflowfactory/resource_workflow_factory.go
```diff
@@ -107,6 +109,18 @@ func (workflows *ResourceWorkflowFactory[R, ResourceDefinition]) resourceDefinit
...
 
+func checkResourceNamespaceAllowed(
+	resourceNamespacedName types.NamespacedName,
+	provider pipelineshub.Provider,
+) error {
+	if len(provider.Spec.AllowedNamespaces) > 0 {
+		if !common.Contains(provider.Spec.AllowedNamespaces, resourceNamespacedName.Namespace) {
+			return fmt.Errorf("resource %s in namespace %s is not allowed by provider %s", resourceNamespacedName.Name, resourceNamespacedName.Namespace, provider.Name)
+		}
+	}
```
TobyPinfold, line 120:
```suggestion
	if len(provider.Spec.AllowedNamespaces) > 0 && !common.Contains(provider.Spec.AllowedNamespaces, resourceNamespacedName.Namespace)  {
		return fmt.Errorf("resource %s in namespace %s is not allowed by provider %s", resourceNamespacedName.Name, resourceNamespacedName.Namespace, provider.Name)
	}
```

TobyPinfold (approved):

peter-hazell:
ok to test

apis/pipelines/v1alpha5/provider_conversion.go
```diff
@@ -17,6 +17,11 @@ func (src *Provider) ConvertTo(dstRaw conversion.Hub) error {
 
 	var beamArgsPatchOps []common.JsonPatchOperation
 
+	remainderFromConvertFrom := ProviderConversionRemainder{}
```
chr12c, line 20: Don't think there's a need to rename this from `remainder`. If anything calling it a remainder-from-convert-from is adding more mental gymnastics. The remainder **must** be from a previous conversion from the hub because it can't be from anywhere else.

chr12c, line 20: This 2 way sharing of the same remainder struct between v1alphaN and v1beta1 is killing me. I can't wait to drop all of these conversions.

chr12c: If i'm reading this correctly, when converting to hub there's no `Image` being fetched.
When converting from hub, there's no `ServiceImage` fetched.

I think this means the conversions To and From shouldn't share the same remainder struct

controllers/pipelines/suite_decoupled_test.go
```diff
@@ -124,6 +124,7 @@ var _ = BeforeSuite(func() {
 			Image: "dummy-image",
 		},
 	}
+	Provider.Spec.AllowedNamespaces = []string{}
```
chr12c, line 127: I think `AllowedNamespaces` should be populated in the decoupled tests

chr12c: Given how the abstract workflow factory has no unit test on the actual workflow construction methods i think at least having it in the decoupled test is the absolute minimum.

TobyPinfold: We were aiming to cover it as much as possible using the unit tests alone as the decouple test suites are already complex enough. I have updated the allowedNamespaces to now explicitly set `default` where all resources are being created.

apis/pipelines/v1alpha5/provider_conversion.go
TobyPinfold: Somewhat unrelated to this issue as this preexists this work. The reason behind this is due to compatibility between v1alpha5 and v1beta1. It's impossible to convert between v1alpha5 to v1beta1 without having prior annotations set for the images. I don't think its worth changing this here. It's only temporary conversion code until v1alpha5 is dropped.

chr12c (approved):

PR #648, opened 2025-05-09
Metrics for run completion event trigger

### LINKED ISSUE #559: Expose basic metrics for Run Completion Event Trigger

The Run Completion Event Trigger does not expose any metrics about its state or incoming requests and outgoing responses. It's a key component of the eventing system and therefore operator managers should be able to monitor it.

Expose standard k8s metrics from the Run Completion Event Trigger pods.

n/a

## Additional context

Similar to https://github.com/sky-uk/kfp-operator/issues/551

- This could be done like this: https://github.com/sky-uk/kfp-operator/pull/547
- Or it could be done us automatic otel instrumentation with just a few lines if no custom metrics are required, this will need to be investigated

chr12c:
Closes #559

- Use go grpc middleware to instrument grpc requests for run completion event trigger.
- Update ConfigMap, Service and Deployment to account for metrics endpoint
- Add ServiceMonitor
- Add app label to Service so ServiceMonitor can match on it
- Add ServiceMonitor for prometheus
- Update default values
- Refactor `ToUrl()` to `ToAddr()` and use builtin `net.JoinHostPort`
- Refactor run completion event trigger's main to use waitgroups and graceful shutdown
- Fix run workflow factory integration test to use static provider


TobyPinfold (approved):

triggers/run-completion-event-trigger/cmd/main.go
```diff
@@ -26,44 +31,93 @@ func main() {
 	}
 
 	ctx := logr.NewContext(context.Background(), logger)
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	go func() {
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, syscall.SIGINT, syscall.SIGTERM)
+		<-c
```
grahamia, line 40:
```suggestion
		channel := make(chan os.Signal, 1)
		signal.Notify(channel, syscall.SIGINT, syscall.SIGTERM)
		<-channel
```

triggers/run-completion-event-trigger/internal/server/metrics_server.go
aidandunlop, line 1: is it too complicated/messy to reuse elements of `provider-service/base/pkg/server/metrics_server.go`?>

chr12c: yes - this metrics server specifically uses the grpc middleware. In general I think the structure and pattern here is far simpler too so not particularly keen on trying to adapt to the old one. As we move on to implement more metric related stuff, I think we might find a good middle ground where we can then make abstractions. But for now I think it's too early to be rigid.

aidandunlop (approved):

triggers/run-completion-event-trigger/cmd/main.go
```diff
@@ -26,44 +31,93 @@ func main() {
...
+		defer waitGroup.Done()
+		logger.Info("Starting metrics server", "addr", metricsSrv.Addr)
+		if err := metricsSrv.ListenAndServe(); err != nil && err != http.ErrServerClosed {
+			logger.Error(err, "Metrics server exited unexpectedly")
+			cancel()
+		}
+	}()
+
+	<-ctx.Done()
+	logger.Info("Context canceled, shutting down servers...")
```
grahamia, line 110:
```suggestion
	logger.Info("Context cancelled, shutting down servers...")
```

grahamia (approved):

PR #668, opened 2025-05-20
Improve run-configuration status feedback when updating

### LINKED ISSUE #667: Improve run-configuration update feedback to client
When a run-configuration is waiting for all runschedules to be in a successful state, it is not clear this is what it is doing. Instead it gives the impression it is hanging without any feedback to the user to explain why.

Populate the message field in the status when updating the runconfiguration so its clear what it is waiting for in order to become successful.

N/A

## Additional context

N/A

TobyPinfold:
Closes #667.

A small and simple improvement, that gives the users some indication what a runconfiguration is waiting for when updating.
It simply populates the currently empty `message` field with a helpful message providing context on what it is doing.

## Tasks

- [ ] ...
- [ ] Documentation updated

### DIFF
--- a/controllers/pipelines/runconfiguration_controller.go
+++ b/controllers/pipelines/runconfiguration_controller.go
@@ -447,7 +447,6 @@ func aggregateState(dependencies []pipelineshub.RunSchedule) (aggState apis.Sync
 		if state == apis.Failed {
 			aggState = apis.Failed
 			message = dependency.Status.Conditions.SynchronizationSucceeded().Message
-			return
 		} else if state != apis.Succeeded {
 			aggState = apis.Updating
 			message = "Waiting for all dependant runschedules to be in a state of Succeeded"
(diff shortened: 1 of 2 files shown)


controllers/pipelines/runconfiguration_controller.go
```diff
@@ -447,9 +447,10 @@ func aggregateState(dependencies []pipelineshub.RunSchedule) (aggState apis.Sync
 		if state == apis.Failed {
 			aggState = apis.Failed
 			message = dependency.Status.Conditions.SynchronizationSucceeded().Message
+			return
 		} else if state != apis.Succeeded {
 			aggState = apis.Updating
-			message = ""
+			message = "Waiting for all dependant runschedules to be succeeded"
```
grahamia, line 453:
```suggestion
			message = "Waiting for all dependant runschedules to be in a state of Succeeded"
```

alexgeorgousis (approved):
As @TobyPinfold pointed out, I'm questioning the logic in this function (I don't fully understand what it's trying to do and whether it's correct). But for now I'm happy with just setting a message to make it easier for users to identify why their RC is stuck in `Updating`.

grahamia (approved):

PR #676, opened 2025-05-23
Support cross-namespace Pipeline Dependencies

### LINKED ISSUE #328: Support Pipeline Dependencies from different namespaces
We'd like to use [Pipeline Dependencies](https://sky-uk.github.io/kfp-operator/docs/examples/pipeline_dependencies/) to import artefacts from Pipelines in a different namespace to the Pipeline they are being imported from. Currently this is limited to both source and destination Pipelines existing in the same namespace.

The ability to configure the namespace of the source pipeline in `runConfigurationRef` and `triggers.runConfigurations`

Assumptions:
- The existing NamespaceName struct can replace the existing string fields to provide the support for namespaced references.
- If no namespace is set then then the namespace of the `RunConfiguration` should be used as the scope when searching for the referenced `RunConfiguration`.
- Security between shared artifacts will be handled by the provider service (VAI / Kubeflow Pipelines) access permissions.

None - current workaround is to deploy all pipelines from the same namespace

**Additional context**
This feature is useful for model ensembling where we can import a foundational model and fine-tuning it for different purposes. It's helpful to organise these Pipelines into separate namespaces.

aidandunlop:
Closes #328.

## Changes
- `RunConfigurationRef.Name` now uses the `common.NamespacedName` struct rather than a string
- `Triggers.RunConfigurations` now uses a list of `common.NamespacedName` rather than a list of strings
- When looking up pipeline dependencies, the `depending_on_runconfiguration_reconciler` will check if a namespace has been specified and look up the base run configuration in that namespace, otherwise it will use the namespace of the current run configuration
- When determining which dependant RunConfigurations to reconcile (and therefore trigger runs for), look across all namespaces, rather than just the namespace of the base RunConfiguration.

## Tasks

- [x] QA
- [x] Documentation updated


apis/pipelines/v1alpha5/run_types.go
alexgeorgousis, line 1: Why is the change also in `v1alpha5` and `v1alpha6`? We normally just make changes to the latest version right?

controllers/pipelines/depending_on_runconfiguration_reconciler.go
```diff
@@ -111,7 +122,13 @@ func (dr DependingOnRunConfigurationReconciler[R]) getIgnoreNotFound(ctx context
 
 func (dr DependingOnRunConfigurationReconciler[R]) setupWithManager(mgr ctrl.Manager, controllerBuilder *builder.Builder, resource client.Object, reconciliationRequestsForPipeline handler.MapFunc) (*builder.Builder, error) {
 	if err := mgr.GetFieldIndexer().IndexField(context.Background(), resource, rcRefField, func(rawObj client.Object) []string {
-		return rawObj.(R).GetReferencedRCs()
+		rc, _ := apis.MapErr(rawObj.(R).GetReferencedRCs(), func(rc common.NamespacedName) (string, error) {
```
chr12c, line 125: `rcnsn` instead of `rc`?

aidandunlop: addressed in f3eea15 - outer rc is now `rcNamespacedNames` and inner can just be `nsn` - I think `rcnsn` is a bit too long

chr12c (approved):

PR #688, opened 2025-05-30
Add metrics to provider server crud routes

### LINKED ISSUE #421: Provider API monitoring - custom metrics
## Overview

After #376 all provider logic will be encapsulated in a live service. So, as with any service, we want users to be able to monitor it.

We need to:
1. Consider which metrics would be valuable for the service to expose
2. Expose the metrics we've chosen

## Chosen Metrics

- [ ] API response status codes (successes and failures)
- [ ] CPU/Memory usage
- [ ] Latency
- [ ] Anything else Provider specific?

## Acceptance Criteria

- [ ] Our chosen metrics are being exported by the provider API

## Notes
We should produce metrics for CRUD operations within Provider services. 

A metric should be incremented upon each CRUD operation within the Provider. This should enable more visibility of operator usage, improve debugging, and enable automatic alerting. This would also allow for correlation and alerting between [Controller’s Reconcile requests](https://github.com/sky-uk/kfp-operator/issues/470) and the Provider service CRUD operations

chr12c:
Closes #421 

Add per-route metrics to all of the provider service's CRUD routes. Implementation uses the `otelhttp` middleware's metrics and injects a `http_route` label that identifies the route. Any `id`s will be ignored to keep the label cardinality under control.

### DIFF
--- a/provider-service/base/go.mod
+++ b/provider-service/base/go.mod
@@ -17,6 +17,7 @@ require (
 	github.com/sky-uk/kfp-operator/common v0.0.0-00010101000000-000000000000
 	github.com/spf13/viper v1.19.0
 	github.com/stretchr/testify v1.10.0
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.49.0
 	go.opentelemetry.io/otel v1.35.0
 	go.opentelemetry.io/otel/exporters/prometheus v0.57.0
 	go.opentelemetry.io/otel/sdk v1.35.0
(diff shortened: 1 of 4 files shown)


chr12c:
`go.opentelemetry.io/otel/sdk/metric/metrictest` is introduced in `go.opentelemetry.io/otel/sdk/metric@v1.36.0` and requires at least golang 1.23.x
Skipping unit tests due to that

peter-hazell (approved):

PR #689, opened 2025-05-30
Training time in run completion events

### LINKED ISSUE #654: Expose `trained-at` timestamp on run completion events

It would be good for run completion events to expose a field that holds the time a run was actually triggered / when training took place. This could then provide a better monitoring and alerting experience for users. 

Add a field `trained-at` to run completion event struct. Updated provider integrations to populate this field.

The VAI Provider already exposes a `"timestamp": "2025-05-13T08:49:27.902834984Z",` field within the message via log router, so this could be used.

Would make sense to provide both `startTime` and `endTime` in the event, so that users can use either field

N/A

## Additional context

N/A

TobyPinfold:
Closes #654 

Adds two optional fields to run completion events, `training.startTime` and `training.endTime` both are RFC3339 format timestamps.

## Tasks

- [x] Tests updated
- [x] Documentation updated


argo/common/run_completion.go
```diff
@@ -61,6 +72,7 @@ type RunCompletionEventData struct {
 	ServingModelArtifacts []Artifact          `json:"servingModelArtifacts"`
 	PipelineComponents    []PipelineComponent `json:"pipelineComponents"`
 	Provider              string              `json:"provider"`
+	Training              *Training           `json:"training,omitempty"`
```
aidandunlop, line 75: this is a bit pedantic but I don't think we should use the term `training` - these are just events of a run of a pipeline, a pipeline might not include any training (i.e. could just be a data ingestion pipeline)

could we not just add them directly to the root of the run completion event?

alexgeorgousis: Good point, I was trying to think of a better name than `training` but moving them to the top makes even more sense 💯

chr12c: The confusing thing about `training` is that would mean everything **not** nested in there has nothing to do with training - which wouldn't be true. Moving to top level sounds sensible unless there's a chance of some other type of start/end time

provider-service/kfp/internal/mocks/mock_kfp_api.go
```diff
@@ -6,8 +6,11 @@ import (
 	"context"
 	"github.com/sky-uk/kfp-operator/argo/common"
 	"github.com/sky-uk/kfp-operator/provider-service/kfp/internal/client/resource"
+	"time"
 )
 
+var StaticTime = time.Now().UTC().Round(0)
```
chr12c, line 12: Why does it need to be rounded?

controllers/webhook/runcompletion_event_trigger.go
```diff
@@ -84,11 +85,31 @@ func RunCompletionEventToProto(event common.RunCompletionEvent) (*pb.RunCompleti
 		ServingModelArtifacts: artifactToProto(event.ServingModelArtifacts),
 		Artifacts:             artifactToProto(event.Artifacts),
 		Status:                statusToProto(event.Status),
+		Training:              trainingToProto(event.Training),
 	}
 
 	return &runCompletionEvent, nil
```
chr12c, line 91: Come to think of it, why does this func need to return a ptr? Users can convert to a ptr if they need to

chr12c, line 91: Come to think of it, why does this func need to return a ptr when no mutation is happening? Users can convert to a ptr if they need to

chr12c, line 91: Come to think of it, why does this func need to return a ptr when no mutation is happening and it's non-optional? Users can convert to a ptr if they need to

provider-service/vai/internal/runcompletion/event_flow.go
```diff
@@ -215,6 +215,19 @@ func (vef *EventFlow) toRunCompletionEventData(ctx context.Context, job *aiplatf
 		Namespace: job.Labels[label.RunConfigurationNamespace],
 	}
 
+	training := &common.Training{}
+	if startTime := job.StartTime; startTime != nil && !startTime.AsTime().IsZero() {
+		st := startTime.AsTime().UTC()
+		training.StartTime = &st
+	}
+	if endTime := job.EndTime; endTime != nil && !endTime.AsTime().IsZero() {
+		et := endTime.AsTime().UTC()
```
chr12c, line 224:
```suggestion
		st := startTime.AsTime()
		training.StartTime = &st
	}
	if endTime := job.EndTime; endTime != nil && !endTime.AsTime().IsZero() {
		et := endTime.AsTime()
```

triggers/run-completion-event-trigger/internal/converters/converters.go
```diff
@@ -22,6 +22,19 @@ func ProtoRunCompletionToCommon(protoRunCompletion *pb.RunCompletionEvent) (comm
...
 
+	var training *common.Training
+	if protoRunCompletion.Training != nil {
+		training = &common.Training{}
+		if protoRunCompletion.Training.StartTime != nil {
+			st := protoRunCompletion.Training.StartTime.AsTime().UTC()
+			training.StartTime = &st
+		}
+		if protoRunCompletion.Training.EndTime != nil {
+			et := protoRunCompletion.Training.EndTime.AsTime().UTC()
```
chr12c, line 33:
```suggestion
			st := protoRunCompletion.Training.StartTime.AsTime()
			training.StartTime = &st
		}
		if protoRunCompletion.Training.EndTime != nil {
			et := protoRunCompletion.Training.EndTime.AsTime()
```

controllers/webhook/runcompletion_event_trigger.go
TobyPinfold: it is optional

provider-service/kfp/internal/mocks/mock_kfp_api.go
TobyPinfold: each time has its own monotonic clock reading, which is slightly different based on the var, so test equality throws up a mismatch even though they are the same time

TobyPinfold: I've added a comment for explanation for future viewing

controllers/pipelines/state_handler.go
```diff
@@ -241,7 +241,7 @@ func (st StateHandler[R]) onSucceededOrFailed(
 
 		targetState = apis.Creating
 	} else {
-		logger.Info("providerId exists, updating")
+		logger.V(2).Info("providerId exists, updating")
```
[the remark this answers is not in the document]
TobyPinfold: done

triggers/run-completion-event-trigger/internal/converters/converters_unit_test.go
```diff
@@ -26,6 +28,10 @@ var _ = Context("ProtoRunCompletionToCommon", func() {
 					Name:     "some-name",
 				},
 			}
+
+			timeNow := time.Now().UTC()
```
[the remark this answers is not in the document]
TobyPinfold: done

argo/common/run_completion.go
TobyPinfold: mived this to the top level as just `runStartTime` `runEndTime`

alexgeorgousis (approved):

PR #690, opened 2025-05-30
Replace apis.utils with lo(dash)

### LINKED ISSUE #629: Replace generic functional functions with lo(dash)
`apis/utils.go` is just a bunch of functional functions that has already been implemented and tested by dedicated open source projects, so there is no need to reinvent the wheel and maintain our own tests

Replace `apis/utils.go` with [lo](https://github.com/samber/lo) (lodash-like), which should have everything required.

A clear and concise description of any alternative solutions or features you've considered, if any.

## Additional context

Any context, screenshots etc. about the feature request that you think is relevant.

chr12c:
Closes #629

Replace `apis.utils` with `lo`

### DIFF
--- a/apis/conditions.go
+++ b/apis/conditions.go
@@ -28,7 +28,10 @@ func ConditionStatusForSynchronizationState(state SynchronizationState) metav1.C
 type Conditions []metav1.Condition
 
 func (conditions Conditions) SynchronizationSucceeded() metav1.Condition {
-	return conditions.ToMap()[ConditionTypes.SynchronizationSucceeded]
+	typeToCondition := lo.Associate(conditions, func(condition metav1.Condition) (string, metav1.Condition) {
+		return condition.Type, condition
+	})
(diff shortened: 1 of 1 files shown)


controllers/pipelines/internal/workflowfactory/pipeline_workflow_factory.go
```diff
@@ -18,9 +19,15 @@ type PipelineParamsCreator struct{}
...
-	return apis.Find(provider.Spec.Frameworks, func(framework pipelineshub.Framework) bool {
+	framework, ok := lo.Find(provider.Spec.Frameworks, func(framework pipelineshub.Framework) bool {
 		return framework.Name == requestedFramework
 	})
+
+	if !ok {
+		return nil, false
+	}
+
+	return &framework, true
```
chr12c, line 30: `apis.Find` returns a ptr (very odd api), but since `lo.Find` is more what you'd expect from a `Find`, it does not return a ptr so this is the extra step.

TobyPinfold (approved):

PR #703, opened 2025-06-05
Update docs: `framework: type` to `framework:name`

aidandunlop:
https://github.com/sky-uk/kfp-operator/pull/611 changed `frameworks.type` -> `framework.name`, but not all of the docs were updated accordingly. This PR fixes the docs and examples to use `name`.

Also fixes the kfpsdk compiler integration-test to run locally if access to temp directory is restricted

## Tasks

- [x] QA
- [x] Documentation updated

### DIFF
--- a/compilers/kfp-sdk/Makefile
+++ b/compilers/kfp-sdk/Makefile
@@ -23,7 +23,7 @@ test: build ## Run unit tests
 
 integration-test: docker-build ## Build and run integration tests
 	$(MAKE) -C ../../docs-gen/includes/master/kfpsdk-quickstart docker-build
-	$(eval TMP := $(shell mktemp -d ./docker-share.XXXXXX))
-	docker run -v $(TMP):/shared ${IMG}:${VERSION} /shared
-	docker run -v $(TMP):/shared -v ./integration:/pipeline --entrypoint /shared/compile.sh kfp-operator-kfpsdk-quickstart:${VERSION} --pipeline_config /pipeline/pipeline.yaml  --provider_config /pipeline/provider.json --output_file /shared/out.json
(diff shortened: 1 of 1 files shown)


TobyPinfold (approved):

grahamia (approved):

compilers/kfp-sdk/Makefile
```diff
@@ -23,6 +23,7 @@ test: build ## Run unit tests
 
 integration-test: docker-build ## Build and run integration tests
 	$(MAKE) -C ../../docs-gen/includes/master/kfpsdk-quickstart docker-build
-	$(eval TMP := $(shell mktemp -d))
+	$(eval TMP := $(shell mktemp -d ./docker-share.XXXXXX))
```
PeterFras, line 26: I experienced similar problem with temp directory on local and running docker with current user and group helped:
```sh
docker run --user $(id -u):$(id -g)
```

compilers/tfx/acceptance/pipeline_conf.yaml
```diff
@@ -1,7 +1,7 @@
 name: namespace/test
 image: test-pipeline
 framework:
-  type: tfx
+  name: tfx
```
PeterFras, line 4: Is the value from this field used/tested anywhere?

aidandunlop: it's not but it's also not used by the compiler itself, it's just passed in

compilers/kfp-sdk/Makefile
aidandunlop: yeah I think I prefer this, added in dfbc90aabc4b60a998557db67c31c91ea536e5c9

PR #736, opened 2025-06-11
Provider webhook sink send events metrics

### LINKED ISSUE #735: Expose provider service webhook response metrics

For better observability of [KFP Operator's eventing journey](https://sky-uk.github.io/kfp-operator/docs/getting-started/introduction/#architecture-overview), we want to have metrics exposed that allow operators to monitor the state of the eventing journey. More specifically, we want to expose the status of the response from the webhook after the provider service has sent the run completion event. The webhook determines if the run completion event was valid, and so this metric should allow operators to collect and analyse the proportion of successful, failed, and discarded events.

This can then be correlated with metrics from Argo's Event Bus to analyse the state of the entire eventing journey.

Expose a custom metric in the provider service that counts webhook responses and their state (successful, failed, discarded, etc)

n/a

## Additional context

The eventing journey starts with events produced when runs complete on external providers (e.g. VAI) and ends with the Argo Events Event Bus where Run Completion Events are published.

TobyPinfold:
Closes #735 

This adds a custom metric to provider services, exposing information on responses to events sent to the controller webhook. 

## Tasks

- [ ] ...
- [ ] Documentation updated


provider-service/base/pkg/streams/sinks/webhook_sink.go
```diff
@@ -20,53 +36,137 @@ type WebhookSink struct {
...
+func (ows ObservedWebhookSink) OnEmpty(ctx context.Context) {
+	logger := common.LoggerFromContext(ctx)
+	logger.Info("webhook sink received empty message")
+	ows.sendEventsCounter.Add(ctx, 1, metric.WithAttributes(attribute.String(sendEventsMetricResultKey, Discarded.String())))
+	ows.delegate.OnEmpty(ctx)
+}
+
+func (ows ObservedWebhookSink) OnError(ctx context.Context, message StreamMessage[*common.RunCompletionEventData]) {
+	logger := common.LoggerFromContext(ctx)
+	jsonMessage, _ := json.Marshal(message.Message)
```
chr12c, line 145: `json.Marshal`'s error channel is not being handled. Similarly to `OnRecoverableFailure` and `OnUnrecoverableFailure`

chr12c: Yea I understand that nothing is done if marshalling fails, but I think there's still value in using the error in the logging message to give more context to users.

PR #767, opened 2025-06-23
Fix makefile test execution

chr12c:
Closes # .

## Tasks

- [ ] ...
- [ ] Documentation updated


Makefile
```diff
@@ -76,23 +78,19 @@ integration-test-down: ## Tear down the minikube cluster
...
 
-test: fmt vet unit-test decoupled-test ## Run all tests
+functional-test: ## Run functional tests
+	$(MAKE) -C triggers/run-completion-event-trigger functional-test
 
-test-argo: ## Run all tests for argo
-	$(MAKE) -C argo/common test
+test: fmt vet unit-test decoupled-test functional-test ## Run all tests
+	# TODO: after integration tests can run on CI, run provider-service as part
+	# of integration-test
```
TobyPinfold, line 87:
```suggestion
	# TODO: after integration tests can run on CI, run provider-service as part
	# of integration-test
```

this comment doesn't make much sense to me, surely this `test` is running as part of the CI process, so it will be running the provider tests?

chr12c: for now, `test` will run provider service integration test in addition to whatever it already runs

TobyPinfold (approved):

PR #772, opened 2025-06-25
Provider validation webhook

### LINKED ISSUE #642: Add kubebuilder validation webhooks to provider and pipeline resources

raised as a result of conversation - https://github.com/sky-uk/kfp-operator/pull/611#discussion_r2071612392

Currently feedback of invalid provider configuration or pipeline configuration in relation to framework setup isn't given till pipeline workflow submission time, it would be a much better user experience if validation was carried out earlier on apply then feedback would be straight away rather than having to look at resource statuses for failure.

Pipeline and provider resources should have validation webhooks added.

Pipeline should be validated to ensure the framework set is allowed by the provider being used by that pipeline. Applying an invalid pipeline should fail with message.

Provider validation should be added to ensure that if a framework is removed then no pipelines exist for that provider that use that framework, change should fail with error if is used.

## Additional context

ie if provider (vai/vai) had the following configuration:
```
spec:
  serviceImage: docker-reg/kfp-operator-vai-provider-service:latest
  serviceAccount: kfp-operator-vai
  frameworks:
    - name: tfx
      image: docker-reg/kfp-operator-tfx-compiler:latest
      patches:
        - patch: '[{"op":"add","path":"/framework/parameters/beamArgs/0","value":somepatch}]'
          type: json
```
and you tried to apply pipeline with (:
```
spec:
  image: docker-reg/kfp-operator-quickstart:latest
  provider: vai/vai
  framework:
    type: kfpsdk
    parameters:
      components: pipeline.create_components
```
Then it should instantly fail on apply. Same other way round if a user of vai provider had a pipeline deployed using tfx framework an error would be thrown on trying to apply an update to the provider removing the tfx framework.

chr12c:
Closes #642

## Tasks

- [ ] ...
- [ ] Documentation updated

### DIFF
--- a/apis/pipelines/hub/provider_types.go
+++ b/apis/pipelines/hub/provider_types.go
@@ -1,6 +1,7 @@
 package v1beta1
 
 import (
+	"github.com/samber/lo"
 	"github.com/sky-uk/kfp-operator/apis"
 	"github.com/sky-uk/kfp-operator/argo/common"
 	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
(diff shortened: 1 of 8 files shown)


TobyPinfold:
Decided this is not required for #642, as the pipeline validation covers the  user operations, the provider resource is deemed the authoritative resource that decided what is / isn't allowed framework wise, and should be able to remove and edit frameworks as required without checking user resources.

PR #784, opened 2025-07-01
Doc Bug: migration guide to v1beta1

grahamia:
Pointed out by a user.

[Migration guide](https://sky-uk.github.io/kfp-operator/docs/reference/resources/migration/v1beta1/) had frameworks rather than framework in the v1beta1 pipeline example

This needed correcting. Think this was missed in original fix as this change was merged at similar time.

- [X] correct pipeline resource examples in migration guide


grahamia:
OK to test

grahamia:
OK to test

PR #813, opened 2025-07-30
Update VAI Label Generation to not trim incorrectly

### LINKED ISSUE #812: VAI label generator is trimming labels incorrectly
**Describe the bug**
The latest implementation of label generation has added trimming to the keys and values of labels incorrectly for VAI. It is wrongly combining the length of both to be less than 64 characters. This limit should apply separately to each, not as a combined total.

It is also trimming with `...` being added which is unsupported.

**To Reproduce**
Deploy a resource which has a combined key, value pair longer than 64 characters.

**Expected behavior**
The limit should be applies separately to both key and value, and the sanitisation should trim them accordingly with no added `...`

**Version and configuration**
If applicable, add configuration details to help explain your problem.
Master Version.

**Logs**
```
Label keys must start with a letter or number, must be less than 64 characters in length,
and must be less that 128 bytes in length when encoded in UTF-8. Label values
must be less than 64 characters in length, and must be less that 128 bytes in
length when encoded in UTF-8."
```
**Additional context**

TobyPinfold:
Closes #812 

- Stops trimming labels incorrectly, instead this will now trim the keys and value to 64 characters each, instead of a combined amount.

- It no longer adds any ellipses to trimmed strings.

## Tasks
- [x] Tests updated


chr12c (approved):

aidandunlop:
Think it might be worth adding something on the end to indicate truncation, as two pipelines of a similar name could end up being truncated to the same thing

aidandunlop (approved):

aidandunlop (approved):

TobyPinfold:
ok to test

TobyPinfold:
ok to test

PR #866, opened 2025-09-11
Support optional run parameters

### LINKED ISSUE #426: Make runtime parameters optional for unresolved references

We want to be able to reference an artifact produced by a previous run of the same pipeline.

Currently, this can be achieved using [pipeline dependencies](https://sky-uk.github.io/kfp-operator/docs/examples/pipeline_dependencies/):
```yaml
kind: RunConfiguration
metadata:
  name: recursive-rc
spec:
  run:
    pipeline: ...
    artifacts:
    - name: examples
      path: ...
    runtimeParameters:
    - name: examples_location
      valueFrom:
        runConfigurationRef:
          name: recursive-rc
          outputArtifact: examples
```

Unfortunately, referencing an output artifact from the same RunConfiguration as a runtime parameter requires the artifact to exist before it can be referenced. This creates a challenge for the first pipeline run, in which the RunConfiguration must be deployed with only the artifact definition.

We propose a way of marking some runtime parameters as optional, to handle cases where the artifact might not exist yet. Example:
```yaml
...
    runtimeParameters:
    - name: examples_location
      valueFrom:
        runConfigurationRef:
          name: recursive-rc
          outputArtifact: examples
          optional: true
```

This would allow a Run or RunSchedule to be created even if some optional references aren't resolved, omitting those parameters. The pipeline would then detect if the parameter was passed and decide on the appropriate action (e.g., resume if an artifact is passed).

**Additional context**
Referencing an output artifact from a previous pipeline run in a current run as a runtime parameter is part of work to explore maintaining state between model training runs. This can be used to resume training or other use cases where new data becomes available at a later time, but where we want to avoid reprocessing old data.

aidandunlop:
Closes #426 

This PR introduces an `optional` field to `RunConfigurationRef` such that users can reference an output artifact before it exists. If an artifact does not exist when the RunConfiguration is applied, and the parameter is marked as optional, then the RunConfiguration will now trigger the run, without the parameter applied. This is particularly useful when a RunConfiguration references itself, as the first time the RunConfiguration is applied, the artifact won't exist.

Also noticed that the docs for RunSchedules are wrong, they do not support ValueFrom parameters directly, so have updated them accordingly.

## Tasks
- [x] Manual QA 
- [x] Documentation updated


controllers/pipelines/run_controller.go
```diff
@@ -87,19 +90,30 @@ func (r *RunReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
...
 	}
 
-	// Never change after being set
-	if run.Status.Dependencies.Pipeline.Version == "" || run.Spec.HasUnmetDependencies(run.Status.Dependencies) {
-		if hasChanged, err := r.handleDependentRuns(ctx, run); hasChanged || err != nil {
+	_, unresolvedOptParams, err := run.Spec.ResolveParameters(run.Status.Dependencies)
+
+	if err != nil {
+		message := fmt.Sprintf("Unable to resolve parameters: %v", err)
+		updateStatus := From(run.Status.Status).WithSyncStateCondition(apis.Succeeded, metav1.Now(), message)
```
chr12c, line 97: Correct me if I'm wrong but I think it's being used like a log/event, which isn't necessary because you've decided to record the event later anyway.

chr12c: 9ab0c925f12151e5b9cea15f3d9d5bce3a7f1868

apis/pipelines/hub/run_types.go
```diff
@@ -92,7 +110,12 @@ func cmpParameters(p1, p2 Parameter) bool {
 		return p1.ValueFrom.RunConfigurationRef.Name.Name < p2.ValueFrom.RunConfigurationRef.Name.Name
 	}
 
-	return p1.ValueFrom.RunConfigurationRef.OutputArtifact < p2.ValueFrom.RunConfigurationRef.OutputArtifact
+	if p1.ValueFrom.RunConfigurationRef.OutputArtifact != p2.ValueFrom.RunConfigurationRef.OutputArtifact {
+		return p1.ValueFrom.RunConfigurationRef.OutputArtifact < p2.ValueFrom.RunConfigurationRef.OutputArtifact
+	}
+
+	return !p1.ValueFrom.RunConfigurationRef.Optional &&
+		p1.ValueFrom.RunConfigurationRef.Optional != p2.ValueFrom.RunConfigurationRef.Optional
```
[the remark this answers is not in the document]
chr12c: very true cfcab684494b8d67b4d8b145978a18da7ef8b15b

apis/pipelines/hub/run_types.go
```diff
@@ -40,14 +42,22 @@ type RunSpec struct {
...
 }
 
-func (runSpec *RunSpec) ResolveParameters(dependencies Dependencies) ([]apis.NamedValue, error) {
-	return apis.MapErr(runSpec.Parameters, func(p Parameter) (apis.NamedValue, error) {
+// ResolveParameters compares Parameters with input status Dependencies.
+// RunConfigurationRefs are checked against the corresponding status dependency:
+// A match implies resolved and a no-match implies dependency artifact has not been published.
+// Unless the RunConfigurationRef has been marked as Optional, this will error.
+// RunConfigurationRefs marked as Optional are considered unresolved to allow for
+// use cases such as recursive dependencies or when dependency is not always required.
```
TobyPinfold, line 50:
```suggestion
// ResolveParameters validates Parameters against dependencies, erroring on unresolved required refs while
// allowing unresolved optional ones for use cases such as recursive dependencies or where a dependency is not // immediately available.
```

PR #870, opened 2025-09-18
Kubeflow provider v2 eventing

### LINKED ISSUE #783: Kfp Provider (v2) Phase 2: eventing integration placeholder
TODO: after the eventless kfp provider v2 phase 1, work out what subtasks are required to integrate it with eventing to get a fully e2e solution. Populate this issue and/or add sub-issues as needed

TobyPinfold:
Closes #783 

Adds eventing integration to Kubeflow v2 provider setup.

## Tasks

- [ ] ...
- [ ] Documentation updated


provider-service/kfp/Makefile
```diff
@@ -12,6 +13,17 @@ unit-test: ## Unit-test the provider-service
 test: unit-test ## Run all tests for the provider-service
 
 ##@ Build
+generate: protoc-gen-go ## Generate service definitions from protobuf
```
chr12c, line 16: This target it not run as part of the root Makefile. Needs to cascade down here

provider-service/kfp/get-proto.mk
```diff
@@ -0,0 +1,16 @@
...
+	[ -d "$(1)@$(2)" ] || { \
+		git clone --no-checkout --branch $(2) --single-branch https://$(1) $(1)@$(2) && \
+		cd $(1)@$(2) && \
+		git sparse-checkout set --no-cone *.proto && \
+		git checkout tags/$(2) -b master; \
+	}
+endef
+
+protoc-gen-go: ## Download protoc-gen-go locally if necessary.
+	$(call go-install,$(PROJECT_DIR)/bin/protoc-gen-go,google.golang.org/protobuf/cmd/protoc-gen-go@v1.28.1)
```
chr12c, line 15: Can you try the [latest version](https://pkg.go.dev/google.golang.org/protobuf@v1.36.10/cmd/protoc-gen-go)
Since we're making these from scratch, it's worth getting the latest because we're probably going to forget ever upgrading unless things break

provider-service/kfp/get-proto.mk
```diff
@@ -0,0 +1,16 @@
...
+	$(call go-install,$(PROJECT_DIR)/bin/protoc-gen-go-grpc,google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.1.0)
```
chr12c, line 16: Similarly, 1.5.1 for this one
https://pkg.go.dev/google.golang.org/grpc/cmd/protoc-gen-go-grpc@v1.5.1

provider-service/kfp/internal/mocks/metadata_store_service_client.go
```diff
@@ -0,0 +1,88 @@
+package mocks
```
chr12c, line 1: Can add relevant build tag here

pkg/common/run_completion.go
```diff
@@ -78,7 +78,7 @@ func (rced RunCompletionEventData) ToRunCompletionEvent() RunCompletionEvent {
 		RunStartTime:          rced.RunStartTime,
 		RunEndTime:            rced.RunEndTime,
 		ServingModelArtifacts: rced.ServingModelArtifacts,
-		Artifacts:             nil,
+		Artifacts:             nil, // to be populated later
```
chr12c, line 81: I'm surprised to see this. I thought `ServingModelArtifacts` is already deprecated, so `Artifacts` should already be in use.

TobyPinfold: Artifacts is also used, but the point at which they are populated on the event is slightly later than the initial creation of the RCE struct, hence the comment to avoid confusion of it 'Not being used' it is in fact used.

chr12c: Makes sense

provider-service/kfp/internal/provider/label_service_test.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("LabelService", func() {
+	Context("InsertLabelsIntoParameters", func() {
+		When("given a list of labels keys, and a compiled pipeline json bytes", func() {
+			It("inserts the labels with a default value into the compiled pipeline", func() {
+				stubPipelineBytes := []byte(`{}`)
```
chr12c, line 14: I don't think there's a need to stub out these values. We're can directly test the implementation of DefaultLabelService, so just instantiate it using the constructor. Then we can validate the default values for free

provider-service/kfp/internal/runcompletion/event_flow.go
```diff
@@ -0,0 +1,196 @@
...
+	Logger         logr.Logger
+	in             chan StreamMessage[*unstructured.Unstructured]
+	out            chan StreamMessage[*common.RunCompletionEventData]
+	errorOut       chan error
+}
+
+const (
+	pipelineRunIdLabel           = "pipeline/runid"
+	workflowPhaseLabel           = "workflows.argoproj.io/phase"
+	workflowUpdateTriggeredLabel = "pipelines.kubeflow.org/events-published"
```
chr12c, line 30: this const is unused.

provider-service/kfp/internal/client/metadata_store.go
```diff
@@ -0,0 +1,207 @@
...
+	if err != nil {
+		return nil, err
+	}
+	if ctxResp.GetContext() == nil {
+		return nil, fmt.Errorf("context not found for runId: %s", runId)
+	}
+
+	// Fetch executions for the context
+	execResp, err := gms.MetadataStoreServiceClient.GetExecutionsByContext(ctx,
+		&ml_metadata.GetExecutionsByContextRequest{ContextId: ctxResp.GetContext().Id})
```
chr12c, line 102: `ctxResp.GetContext()` is double-fetched (L96)

provider-service/kfp/internal/provider/label_service_test.go
TobyPinfold: what benefit is that providing? I think it just makes the assertion more complicated, the behaviour is what we are testing and this does so, whilst staying as simple as possible 🤷‍♂️

provider-service/kfp/internal/client/kfp_api.go
```diff
@@ -0,0 +1,96 @@
...
+	if run.CreatedAt != nil {
+		runCreateTime := run.CreatedAt.AsTime()
+		resourceReferences.CreatedAt = &runCreateTime
+	}
+
+	if run.FinishedAt != nil {
+		runFinishedTime := run.FinishedAt.AsTime()
+		resourceReferences.FinishedAt = &runFinishedTime
+	} else {
+		// the FinishededAt field is unreliable, so if it's nil we set it to the current time
```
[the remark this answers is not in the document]
TobyPinfold: It either doesn't get set or comes back as epoch 0. This is a workaround, its not the best but the time shouldn't be far off given its based off the workflows anyhow.

provider-service/kfp/internal/client/metadata_store.go
```diff
@@ -0,0 +1,207 @@
...
+				})
+			}
+		}
+	}
+
+	return results, nil
+}
+
+func (gms *GrpcMetadataStore) GetArtifactsForRun(ctx context.Context, runId string) ([]common.PipelineComponent, error) {
+	// 1. Resolve run context
```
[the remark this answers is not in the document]
TobyPinfold: Think I cleared up most and missed this 🫢

provider-service/kfp/internal/runcompletion/event_flow.go
TobyPinfold: It was used within the test suite, moved to the test.

provider-service/kfp/internal/client/metadata_store.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, err
+	}
+	artifactByID := lo.Associate(artsResp.Artifacts, func(a *ml_metadata.Artifact) (int64, *ml_metadata.Artifact) {
+		return a.GetId(), a
+	})
+
+	// Build PipelineComponents from executions and their artifacts
+	pcs := lo.FilterMap(executions, func(exec *ml_metadata.Execution, _ int) (common.PipelineComponent, bool) {
+		pc := common.PipelineComponent{
+			Name:               exec.CustomProperties[DisplayName].GetStringValue(),
```
chr12c, line 145: accessing this DisplayName key can potentially return nil. There's no nil check here

provider-service/kfp/internal/provider/label_service.go
```diff
@@ -0,0 +1,47 @@
...
+	parameterDefaults []byte
+	parameterJsonPath string
+}
+
+func NewDefaultLabelService() (LabelService, error) {
+	paramDef := map[string]any{
+		"isOptional":    true,
+		"parameterType": "STRING",
+		"defaultValue":  "",
+	}
```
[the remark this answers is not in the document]
TobyPinfold: These are the values required within the compiled pipeline. Given the parameters here are representing `labels` the obvious default values and optional fields make sense.

provider-service/kfp/internal/provider/label_service_test.go
chr12c: would sense if this label service had dependencies that can realistically differ in behaviour in the concrete implementation, but that's true in this case because the beavhiour is dictated by the hardcoded values.

provider-service/kfp/internal/runcompletion/event_decoupled_test.go
```diff
@@ -0,0 +1,322 @@
...
+	defaultNamespace             = "default"
+	providerName                 = "kfp"
+	webhookUrl                   = "/operator-webhook"
+	workflowUpdateTriggeredLabel = "pipelines.kubeflow.org/events-published"
+)
+
+var _ = BeforeSuite(func() {
+	ctx := context.Background()
+
+	testEnv := &envtest.Environment{
```
chr12c, line 64: Have a look at this PR https://github.com/sky-uk/kfp-operator/pull/645 where envtest setup was changed. I think it needs to [download the binary assets ](https://github.com/sky-uk/kfp-operator/blob/dce6fda28e82c335c0f405d8a9d6893220a88e35/controllers/pipelines/suite_decoupled_test.go#L52)in case this test is ran first in the CI

chr12c (approved):
LGTM

chr12c:
OK to test

peter-hazell (approved):

PR #946, opened 2025-11-18
Two separate helm charts

### LINKED ISSUE #890: Separate cluster-wide resources from KFP Operator/provider installation

It should be possible to install cluster-wide resources needed for the KFP Operator separately than namespace specific resources, for users that don't have full control over their Kubernetes cluster. 

That way, cluster admins could install CRDs, Cluster Roles etc while Operator managers could install the controller, eventing, and Providers themselves.

Split the Helm chart into two component parts, one for cluster-wide resources, the other with the rest, so that they can be installed separately by different users.

Force those without full control over their Kubernetes cluster to have to rely on cluster admins to look after the operator installation.

chr12c:
Closes #890

Move crds to a separate helm chart, which gives flexibility to users who do not have permissions to deploy their own crds.

## Tasks

- [ ] ...
- [ ] Documentation updated


chr12c:
superseded by https://github.com/sky-uk/kfp-operator/issues/964

PR #956, opened 2025-12-04
fix helm publish OCI destination extraneous path segment

chr12c:
fix helm publish OCI destination to not use an extraneous path segment (kfp-operator) because the registry reference basename should be inferred from the chart's name.


peter-hazell (approved):

alexgeorgousis (approved):
nice one @chr12c

PR #965, opened 2025-12-10
Enforce all eventbuses to run as non root user

chr12c:
Enforce all eventbuses to run as non root user, which aligns with the principle of least privilege.


chr12c:
OK to test

PR #1011, opened 2026-01-29
remove unused table styling from _styles_project.scss

chr12c:
Removes custom table styling that was not being applied due to CSS specificity issues. The custom `.table` styles in `_styles_project.scss` were being completely overridden by Docsy's theme-aware table styles in both light and dark modes due to higher CSS specificity. Docsy uses `[data-bs-theme=dark]` wrappers for dark mode and more specific selectors like `.td-content > table` for light mode, both of which have higher specificity than the plain `.table` selector used in our custom styles. Since the tables are already using Docsy's default styling, there's no need to maintain this unused code.


chr12c:
OK to test

PR #1012, opened 2026-01-30
use helm 'with' to eliminate redundant value path references

### LINKED ISSUE #997: Refactor Helm templates to use with instead of if for conditional YAML blocks
## Background

Currently, our Helm templates use the if conditional pattern when rendering optional YAML blocks. While this works correctly, Helm provides a more idiomatic approach using the with keyword that improves readability and reduces repetition.

## The Problem

When we use if to conditionally render a value and then pipe it through toYaml, we have to repeat the full path to the value:
```
{{- if .Values.manager.argo.securityContext }}
securityContext:
    {{- .Values.manager.argo.securityContext | toYaml | nindent 4 }}
{{- end }}
```
## The Solution

Using with is more idiomatic because it:
1. Reduces repetition - no need to repeat the long value path
2. Improves readability - clearer intent that we're working with that specific value
3. Follows Helm best practices - this is the recommended pattern for this use case

The refactored version:
```
{{- with .Values.manager.argo.securityContext }}
securityContext:
    {{- . | toYaml | nindent 4 }}
{{- end }}
```

Note that inside a with block, . is redefined to refer to the value being checked, so we use . instead of the full path.

chr12c:
Use helm 'with' to eliminate redundant value path references. When conditionally rendering values with toYaml, the same value path was being referenced twice

Closes #997


chr12c:
OK to test

PR #1151, opened 2026-04-16
chore(deps): exclude archived doc versions from dependabot scanning

### LINKED ISSUE #1149: Archived docs triggering dependabot updates
Archived docs, e.g. `0.6.0` and `0.7.0` are triggering updates from dependabot even though the config does not include those directories because security updates will trigger regardless. There are many ways to get around this, but the main point is not having `pyproject.toml` and `poetry.lock` etc. to trigger dependabot.

chr12c:
- Archived doc versions still contain pyproject.toml and poetry.lock from before the uv migration. These trigger unwanted security PRs for dependencies that are not relevant to master code.
- Watch for tfx related library upgrades (master-only)

Closes #1149

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,8 +19,8 @@ updates:
 
   - package-ecosystem: "uv"
     directories:
-    - "/compilers/kfp-sdk"
-    - "/docs-gen/includes/master/kfpsdk-quickstart"
+    - "/compilers/**"
+    - "/docs-gen/includes/master/**"
     schedule:
       interval: "weekly"
     rebase-strategy: "disabled"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,7 +20,7 @@ updates:
   - package-ecosystem: "uv"
     directories:
     - "/compilers/kfp-sdk"
-    - "/docs-gen/includes/*/kfpsdk-quickstart"
+    - "/docs-gen/includes/master/kfpsdk-quickstart"
     schedule:
       interval: "weekly"
     rebase-strategy: "disabled"
@@ -28,3 +28,16 @@ updates:
     - dependency-type: "direct"
     commit-message:
       prefix: "chore(deps)"
+    exclude-paths:
+    - "docs-gen/includes/versions/**"
+
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    rebase-strategy: "disabled"
+    commit-message:
+      prefix: "chore(deps)"
+    exclude-paths:
+    - "docs-gen/includes/versions/**"
+    open-pull-requests-limit: 0


.github/dependabot.yml
```diff
@@ -20,11 +20,24 @@ updates:
...
       interval: "weekly"
     rebase-strategy: "disabled"
     allow:
     - dependency-type: "direct"
     commit-message:
       prefix: "chore(deps)"
+    exclude-paths:
+    - "docs-gen/includes/versions/**"
+
+  - package-ecosystem: "pip"
```
chr12c, line 34: 0.6.0 and 0.7.0 use pip, so we need this block until we eventually delete them.

mdiedricks (approved):

.github/dependabot.yml
```diff
@@ -20,11 +20,24 @@ updates:
   - package-ecosystem: "uv"
     directories:
     - "/compilers/kfp-sdk"
-    - "/docs-gen/includes/*/kfpsdk-quickstart"
+    - "/docs-gen/includes/master/kfpsdk-quickstart"
```
bcarter97: Do we want to include `/docs-gen/includes/master/quickstart` and `/docs-gen/includes/master/quickstart-with-dependencies` as these have uv locks?

.github/dependabot.yml
```diff
@@ -20,11 +20,24 @@ updates:
...
     schedule:
       interval: "weekly"
     rebase-strategy: "disabled"
     allow:
     - dependency-type: "direct"
     commit-message:
       prefix: "chore(deps)"
+    exclude-paths:
+    - "docs-gen/includes/versions/**"
```
bcarter97, line 32: Think this is redundant if the directories only includes `master` and not `versions`

.github/dependabot.yml
```diff
@@ -20,11 +20,24 @@ updates:
...
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    rebase-strategy: "disabled"
+    commit-message:
+      prefix: "chore(deps)"
+    exclude-paths:
+    - "docs-gen/includes/versions/**"
+    open-pull-requests-limit: 0
```
bcarter97:
```suggestion
    open-pull-requests-limit: 0 # ignore pip dependencies for archived doc versions
```

.github/dependabot.yml
chr12c: That's how you'd hope it works, but we know for sure that currently security updates will ignore what's in the directories section.

bcarter97: In that case do security updates respect `exclude-paths`?

chr12c: according to the docs, it should: https://github.blog/changelog/2025-08-26-dependabot-can-now-exclude-automatic-pull-requests-for-manifests-in-selected-subdirectories/

We can observe how it behaves after this is merged.

.github/dependabot.yml
chr12c: 15d63f50ee255adf6bbd9b55d4c91f8ec4d0fd55

we should also be watching for tfx related upgrades

TobyPinfold (approved):

PR #1172, opened 2026-04-20
chore(kfp-sdk compiler): migrate linting from black+flake8 to ruff

chr12c:
This reduces dev dependencies from 6 packages (black, flake8, pycodestyle, pyflakes, mccabe, mypy-extensions) to 1 (ruff).

- Use ruff's recommended rule set (E, F, UP, B, SIM, I) which adds pyupgrade, flake8-bugbear, flake8-simplify and isort on top of the pycodestyle and pyflakes rules previously enforced by flake8.
- Move lint configuration from the separate .flake8 file into pyproject.toml.
- Change the build target to verify formatting (--check) rather than rewriting source files in-place.
- Fix violations introduced by the new rule set.

### DIFF
--- a/compilers/kfp-sdk/.flake8
+++ b/compilers/kfp-sdk/.flake8
@@ -1,4 +1,0 @@
-[flake8]
-exclude = 
-    .venv
-max-line-length = 120
--- a/compilers/kfp-sdk/Makefile
+++ b/compilers/kfp-sdk/Makefile
@@ -7,8 +7,8 @@ lock: ## Update uv.lock after changing pyproject.toml
 
 build: ## Build
 	uv sync --locked --all-extras
-	uv run black .
-	uv run flake8
+	uv run ruff format --check .
+	uv run ruff check .
 	uv build
 
 ##@ Containers
--- a/compilers/kfp-sdk/Makefile
+++ b/compilers/kfp-sdk/Makefile
@@ -21,6 +21,10 @@ DOCKER_BUILD_EXTRA_PARAMS=--build-arg WHEEL_VERSION=${WHEEL_VERSION} --build-arg
 
 ##@ Development
 
+format: ## Format and fix source files in place
+	uv run ruff format .
+	uv run ruff check --fix .
+
 test: build ## Run unit tests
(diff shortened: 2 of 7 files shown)


compilers/kfp-sdk/pyproject.toml
```diff
@@ -27,6 +26,9 @@ packages = ["compiler"]
 version_scheme = "post-release"
 local_scheme = "dirty-tag"
 
+[tool.ruff.lint]
+select = ["E", "F", "UP", "B", "SIM", "I",]
```
bcarter97:
```suggestion
select = ["E", "F", "UP", "B", "SIM", "I"]
```

compilers/kfp-sdk/.flake8
```diff
@@ -1,4 +0,0 @@
-[flake8]
-exclude = 
-    .venv
-max-line-length = 120
```
peter-hazell, line 4: do we not want to set max line length with ruff?

compilers/kfp-sdk/Makefile
```diff
@@ -7,8 +7,8 @@ lock: ## Update uv.lock after changing pyproject.toml
 
 build: ## Build
 	uv sync --locked --all-extras
-	uv run black .
-	uv run flake8
+	uv run ruff format --check .
+	uv run ruff check .
```
peter-hazell, line 11: is it worth having a separate make target like `format` which will format source files in place rather than just check? would make it easier to fix any formatting issues that get picked up by CI

compilers/kfp-sdk/.flake8
chr12c: We will use the default, which happens to be what's the most popular anyway for python (and it's <120)

compilers/kfp-sdk/Makefile
chr12c: 0e71be47ae3d5f4e580f86888e7377ef3e53d6e5

peter-hazell (approved):

bcarter97 (approved):

PR #1183, opened 2026-04-22
chore(docs-gen): inline archived docs

### LINKED ISSUE #1149: Archived docs triggering dependabot updates

chr12c:
Closes #1149

`archive-version` inlines code. In particular it no longer archives files that cause dependabot to consider archived docs as targets for updating. Only applies to future archived docs and not the current `0.6.0` and `0.7.0`.

### DIFF
--- a/docs-gen/Makefile
+++ b/docs-gen/Makefile
@@ -56,28 +56,50 @@ archive-version: ## Archive a version of the documentation. Requires VERSION to 
 		exit 1; \
 	fi
 
-	# Copy existing docs into a versioned directory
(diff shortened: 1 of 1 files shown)


chr12c:
superseded by https://github.com/sky-uk/kfp-operator/pull/1184

PR #1185, opened 2026-04-22
chore(docs-gen): inline code in 0.6.0 and 0.7.0 docs

chr12c:
old python examples removed as files to prevent dependabot version and security updates from triggering on them.

Connected to #1149


chr12c:
superseded by https://github.com/sky-uk/kfp-operator/pull/1184

PR #1241, opened 2026-05-27
fix(kfp provider): pipelineRootStorage determining artifact output location

### LINKED ISSUE #1242: kfp provider not using pipelineRootStorage
**Describe the bug**
Currently the pipelineRootStorage in kfp provider does nothing. It needs to be passed into the arguments when calling the kubeflow server to execute runs/recurring runs. This will make it so that output artifacts can be configured properly.

**To Reproduce**
trigger a run via kfp provider and have pipelineRootStorage configured to gcs. Observe kubeflow workflow pod logs show that it writes output artifacts to minio instead.

**Expected behavior**
Should attempt to write to gcs if configured to gcs, for example

chr12c:
Currently the pipelineRootStorage in kfp provider does nothing. This PR fixes this such that it determines where the output artifacts are created, similar to the vai provider.

Closes https://github.com/sky-uk/kfp-operator/issues/1242

### DIFF
--- a/provider-service/kfp/internal/provider/run_service.go
+++ b/provider-service/kfp/internal/provider/run_service.go
@@ -72,14 +72,13 @@ func (rs DefaultRunService) CreateRun(
 		return "", err
 	}
 
-	namespacedName, err := rd.PipelineName.String()
-	if err != nil {
-		return "", err
-	}
-
 	outputDirectory := ""
 
 	if rs.pipelineRootStorage != "" {
+		namespacedName, err := rd.PipelineName.String()
+		if err != nil {
+			return "", err
+		}
 		outputDirectory = fmt.Sprintf("%s/%s", rs.pipelineRootStorage, namespacedName)
 	}
 
--- a/provider-service/kfp/internal/provider/run_service.go
+++ b/provider-service/kfp/internal/provider/run_service.go
@@ -26,13 +26,15 @@ type RunService interface {
 }
 
 type DefaultRunService struct {
-	client         client.RunServiceClient
-	labelGenerator label.LabelGen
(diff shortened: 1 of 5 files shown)


provider-service/kfp/internal/provider/run_service.go
```diff
@@ -80,7 +94,8 @@ func (rs DefaultRunService) CreateRun(
 				},
 			},
 			RuntimeConfig: &go_client.RuntimeConfig{
-				Parameters: runParameters,
+				Parameters:   runParameters,
+				PipelineRoot: outputDirectory,
```
peter-hazell, line 97: This should be added to `recurring_run_service` as well for scheduled runs

provider-service/kfp/internal/provider/run_service.go
```diff
@@ -69,6 +72,17 @@ func (rs DefaultRunService) CreateRun(
 		return "", err
 	}
 
+	namespacedName, err := rd.PipelineName.String()
+	if err != nil {
+		return "", err
+	}
```
peter-hazell: can move this into the if statement as it's not needed outside?

chr12c: 480b2eedda0a5f3cca11ad9a159359b02a817411

provider-service/kfp/internal/provider/run_service.go
chr12c: 92b9e8d2f3cacaed4c256dde6bcd4239541bfc38

chr12c:
OK to test

peter-hazell (approved):

TobyPinfold (approved):

Issue #730, opened 2025-06-10
Add Support for Kubeflow Pipelines v2


alexgeorgousis:
# Overview

KFP v1 has been [officially deprecated](https://github.com/kubeflow/pipelines/discussions/11781), so we want to support the newer version, KFP v2. Additionally, the ability of the compiler to compile to v1 [is broken](https://github.com/sky-uk/kfp-operator/issues/389). Given those two facts, we want to drop existing support for KFP v1 and then introduce support for KFP v2.

## Current Setup

The KFP Operator currently includes:
- A Python compiler for TFX pipelines which is parameterised to use either v1 or v2 of the [KFP SDK](https://kubeflow-pipelines.readthedocs.io/en/stable/).
    - v1 compiles to a YAML representation of the Pipeline (supported by KFP v1).
        - This is currently broken (see #389).
        - Also KFP v1 is [officially deprecated](https://github.com/kubeflow/pipelines/discussions/11781).
    - v2 compiles to a JSON representation of the Piepline (supported by KFP v2).
- A Provider Service for VAI which uses the v2 KFP API to upload the v2 JSON representation to VAI (VAI uses KFP under the hood).
- A Provider Service for KFP which uses the v1 KFP API to upload the v1 YAML representation to KFP.
    - Obviously doesn't work since the v1 compiler doesn't work.

## Desired Setup

- A compiler that only uses KFP v2.
- (Already exists) A VAI Provider Service.
- (Needs to be added) A KFP v2 Provider Service.

# Strategy

Our strategy is to:
1. Drop support for KFP v1 by removing that option from the compiler and dropping the existing KFP Provider Service.
2. Introduce a KFP v2 Provider Service.

# Useful Links

- [Official KFP v1 -> v2 migration docs](https://www.kubeflow.org/docs/components/pipelines/user-guides/migration/)

PR #1248, opened 2026-06-04
feat(TFX): Add patch image to support TFX pipelines running on kubeflow pipelines v2

TobyPinfold:
links to #730 

Long running issue with TFX being in maintenance mode and its support for kubeflow pipelines v2 is not fully functional. This provides a short term fix for most issues encountered such that users can run TFX on kubeflow via the KFP Operator.  

## Tasks

- [x] Documentation updated


provider-service/kfp/tfx-patch/patch_tfx_kfp_v2.py
```diff
@@ -0,0 +1,354 @@
...
+)
+_PATH_UTILS = "tfx.utils.path_utils"
+
+
+# ---------------------------------------------------------------------------
+# Patch data model
+# ---------------------------------------------------------------------------
+
+@dataclass(frozen=True)
+class Patch:
```
chr12c: Have you looked into using an import hook + shim module rather than string editing? It's better that way because it can survive formatting changes and you get to depend on the symbols instead of exact strings. The patch would also be an importable module that you can directly unit test, so it's far more robust.

peter-hazell: discussed with Toby last week and this got added https://github.com/sky-uk/kfp-operator/pull/1248/commits/283ac5b9b95286a8e3288413a768ec822f0692e3

we should probably discuss/decide which way forward we want to take because at the moment this PR has 2 separate approachs

peter-hazell: readme also needs updating to cover this as it just covers the string editing patch currently

provider-service/kfp/tfx-patch/README.md
```diff
@@ -0,0 +1,61 @@
+# TFX Patch Image
+
+A lightweight `scratch`-based image containing a single Python script (`patch_tfx_kfp_v2.py`) that fixes five TFX runtime incompatibilities with the OSS KFP v2 driver/launcher.
```
QLyine: I might be wrong, but README seems to be different from the actual Dockerfile (implementation) that is using install_shim.py.

```sh
# Install the shim (import hook approach — no source file modification)
COPY tfx_kfp_v2_shim/ /tmp/tfx_kfp_v2_shim/
RUN python /tmp/tfx_kfp_v2_shim/install_shim.py && rm -rf /tmp/tfx_kfp_v2_shim
```

provider-service/kfp/tfx-patch/tfx_kfp_v2_shim/_patches.py
```diff
@@ -0,0 +1,376 @@
...
+                    log.debug("Inferred %s from keys %s", tfx_title, all_keys)
+                    if kind == "schema_title":
+                        artifact_pb.type.schema_title = tfx_title
+                    break
+
+        # Guard null instance_schema so the original won't crash (patch 2).
+        if kind == "instance_schema" and not type_schema.instance_schema:
+            artifact_pb.type.instance_schema = "title: ''"
+
+        # Debug: log the type schema being processed
```
QLyine: I imagine this is intended ? To leave this debug here ? Or should it be log.debug ?

provider-service/kfp/tfx-patch/tfx_kfp_v2_shim/_patches.py
```diff
@@ -0,0 +1,376 @@
...
+    """Delete zero-byte blobs ending with ``/`` under *uri*."""
+    try:
+        from google.cloud import storage as gcs_storage
+
+        path = uri[5:]  # strip "gs://"
+        bucket_name, _, prefix = path.partition("/")
+        if prefix and not prefix.endswith("/"):
+            prefix += "/"
+
+        client = gcs_storage.Client()
```
QLyine: I guess we are creating the client every time this function is called.

Should the client be lazily initialized or cached instead? Or perhaps the code could be rearranged differently altogether, if that makes sense. Or maybe it is fine as is.

provider-service/kfp/tfx-patch/tfx_kfp_v2_shim/_patches.py
```diff
@@ -0,0 +1,376 @@
...
+    def _flat_serving_model_dir(output_uri: str, is_old_artifact: bool = False) -> str:
+        return output_uri
+
+    mod.serving_model_dir = _flat_serving_model_dir
+    log.info("Patch 5: flattened serving_model_dir")
+
+
+# ── Patch 7: zero-byte directory marker cleanup (run_executor) ─────────────────
+
+def patch_run_executor(mod: ModuleType) -> None:
```
QLyine: Is this not dead code? This is instead done by `_patch_translate_executor_output` in `patch_entrypoint_utils` in `def install(*, vertex_compatible: bool = False) -> None:`

QLyine (changes requested):

provider-service/kfp/tfx-patch/tfx_kfp_v2_shim/_patches.py
```diff
@@ -0,0 +1,376 @@
...
+        path = uri[5:]  # strip "gs://"
+        bucket_name, _, prefix = path.partition("/")
+        if prefix and not prefix.endswith("/"):
+            prefix += "/"
+
+        client = gcs_storage.Client()
+        bucket = client.bucket(bucket_name)
+
+        deleted = 0
+        for blob in bucket.list_blobs(prefix=prefix):
```
QLyine: Minor perf thing. 

If you think it's better this can be done in batch. I am not well informed to know if it's worth it.

```python
  def _delete_gcs_directory_markers(uri: str) -> None:
      """Delete zero-byte blobs ending with ``/`` under *uri*."""
      try:
          from google.cloud import storage as gcs_storage

          path = uri[5:]  # strip "gs://"
          bucket_name, _, prefix = path.partition("/")
          if prefix and not prefix.endswith("/"):
              prefix += "/"
          client = gcs_storage.Client()
          bucket = client.bucket(bucket_name)

          markers = [
              blob for blob in bucket.list_blobs(prefix=prefix)
              if blob.size == 0 and blob.name.endswith("/")
          ]

          if markers:
              with client.batch():
                  for blob in markers:
                      blob.delete()
              log.info(
                  "Patch 7: deleted %d GCS directory marker(s) under %s",
                  len(markers), uri,
              )
      except Exception as exc:
          log.warning(
              "Patch 7: could not clean markers under %s: %s", uri, exc,
          )
```

provider-service/kfp/tfx-patch/README.md
peter-hazell: ☝️ this is because a comment was made suggesting that we implement the shim approach instead of the string patching approach

the readme hasn't been updated

IMO we should agree which approach we want to go with and only include that one on this PR rather than having both. I'm assuming based on discussions that the shim is the 'safer' approach and we should probably go with that

provider-service/kfp/tfx-patch/tfx_kfp_v2_shim/_patches.py
TobyPinfold: I've moved it to a shared client

provider-service/kfp/tfx-patch/tfx_kfp_v2_shim/_patches.py
TobyPinfold: moved to a batch approach

QLyine (approved):

sky-danielsantos (approved):

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG UV_VERSION
+FROM ghcr.io/astral-sh/uv:${UV_VERSION} AS uv
+
+FROM python:3.10.12
+
+COPY --from=uv /uv /bin/uv
+
+COPY pyproject.toml uv.lock ./
```
chr12c: there is no `uv.lock`

provider-service/kfp/tfx-patch/tests/test_shim.py
chr12c, line 1: how and where do these tests get run.

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
```diff
@@ -0,0 +1,22 @@
+ARG UV_VERSION
```
chr12c: where does this file get built

TobyPinfold: it doesn't its really just for testing purposes / demo / example

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
TobyPinfold: added

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
chr12c: If it's for testing purposes then it should be run as part of CI. If it ever breaks (like how it currently is), then it could go under the radar

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+# Install the shim (import hook approach — no source file modification)
+COPY tfx_kfp_v2_shim/ /tmp/tfx_kfp_v2_shim/
+RUN python /tmp/tfx_kfp_v2_shim/install_shim.py && rm -rf /tmp/tfx_kfp_v2_shim
+
+RUN mkdir /data
+RUN wget https://raw.githubusercontent.com/tensorflow/tfx/master/tfx/examples/penguin/data/labelled/penguins_processed.csv -P /data
+
+# Install pipeline code as a package into site-packages so it's importable
+# from the compiler (which runs via compile.sh with its own PYTHONPATH)
+COPY penguin_pipeline/ /usr/local/lib/python3.10/site-packages/penguin_pipeline/
```
chr12c: `penguin-pipeline` is not committed.

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
```diff
@@ -0,0 +1,22 @@
...
+FROM python:3.10.12
+
+COPY --from=uv /uv /bin/uv
+
+COPY pyproject.toml uv.lock ./
+ENV UV_PROJECT_ENVIRONMENT=/usr/local
+RUN uv sync --locked --no-install-project --no-dev
+
+# Install the shim (import hook approach — no source file modification)
+COPY tfx_kfp_v2_shim/ /tmp/tfx_kfp_v2_shim/
```
chr12c: `tfx_kfp_v2_shim` is one level higher

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
chr12c: this is a pattern copied from other dockerfiles/makefiles, but it does not apply here correctly. Where is it getting the value of the argument here

provider-service/kfp/tfx-patch/README.md
```diff
@@ -0,0 +1,61 @@
...
+| 6 | `tfx.v1.orchestration.experimental` | Re-export `KubeflowV2DagRunner` removed by `kfp>=2` |
+| 7 | `kubeflow_v2_entrypoint_utils` | Delete zero-byte GCS directory markers after each executor run |
+
+Patches 1–4 and 6–7 are safe for all environments. Patch 5 changes TFX's model layout and breaks Vertex AI — use `--vertex-compatible` to skip it.
+
+Patches 5 and 7 work around a KFP launcher bug ([kubeflow/pipelines#13476](https://github.com/kubeflow/pipelines/issues/13476)). They become redundant once that is fixed upstream.
+
+## Supported TFX versions
+
+1.15, 1.16, 1.17
```
chr12c, line 30: The test only does 1.15.1, so not all of these versions are proven to be supported.

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
chr12c: 9f15580f2ddc8e0bb3e9f52d051052473ca3a271

provider-service/kfp/tfx-patch/README.md
TobyPinfold: I tested it with all 3 versions.

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
TobyPinfold: I've just removed the test-pipeline entirely, it was purerly for demo purposes, guess the readme will suffice anyhow.

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
TobyPinfold: removed

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
TobyPinfold: removed

provider-service/kfp/tfx-patch/test-pipeline/Dockerfile
TobyPinfold: removed it anyway

provider-service/kfp/tfx-patch/tests/test_shim.py
TobyPinfold: from the makefile.

chr12c: There is a `test` target for it, but there isn't an aggregate target that actually invokes it.

peter-hazell (approved):

provider-service/kfp/tfx-patch/tests/test_shim.py
TobyPinfold: I'm of the opinion it doesn't make sense to have an aggregated target for it, its not a production ready service etc, just a optional patch. It makes sense to test it when changing anything but for automation it would be wasted effort / time

PR #1249, opened 2026-06-04
feat(tfx-compiler): Allow for pipeline framework to determine output pipeline spec version

### LINKED ISSUE #763: Support Schema 2.1 in TFX Compiler output
With kfpsdk framework compiler the output if Schema 2.1, it would be nice if it was possible to produce schema 2.1 output in the TFX compiler.

2.1 output support was added to TFX in version 1.15.1 and you need to set in the `KubeflowV2DagRunnerConfig` init setting `use_pipeline_spec_2_1` to true.

See [code here](https://github.com/tensorflow/tfx/blob/e537507b0c00d45493c50cecd39888092f1b3d79/tfx/orchestration/kubeflow/v2/kubeflow_v2_dag_runner.py#L195)

I think a flag will need to be set in the pipeline resource (could use a specific value within the parameters of the tfx framework) to indicate to the KFP Operator TFX compiler to use the v2 of the config object and set the above flag to true. Cant just be set as only had support added in 1.15.1 tfx

## Additional context
Allowing Schema 2.1 will allow the use of template registry to upload the pipelines to rather than direct to GCS.

TobyPinfold:
Closes #763 

- Adds support for pipeline spec 2.1 schema for tfx pipelines. 
- Provides control per pipeline via `framework` parameter `usePipelineSpec2_1` (defaults to 2.1) 
- Adds KFP provider support for tfx spec (unwrap from VAI specific pipeline shape)
- Updates TFX framework docs to include `usePipelineSpec2_1` as a configuration.
- Decouples compiler tfx version from pipelines and decouples target kfp version. This means pipelines requiring higher kfp versions for some platform features can now do so without having to use a separate compiler
- Adds integration tests with pipelines using python 3.10 - tfx 1.16.0 and tfx 1.17.3

## Tasks

- [x] Documentation updated


provider-service/kfp/internal/provider/provider_test.go
```diff
@@ -497,4 +499,35 @@ var _ = Describe("Provider", func() {
 			})
 		})
 	})
+
+	Context("unwrapTfxPipelineSpec", func() {
+		It("should extract pipelineSpec from wrapper for TFX framework (case-insensitive)", func() {
+			innerSpec := map[string]interface{}{
+				"pipelineInfo": map[string]interface{}{"name": "test-pipeline"},
```
chr12c:
```suggestion
				"pipelineInfo": map[string]any{}{"name": "test-pipeline"},
```
same below

provider-service/kfp/internal/provider/provider.go
```diff
@@ -81,6 +84,30 @@ func NewKfpProvider(config *config.Config) (*KfpProvider, error) {
...
+// Kubeflow Pipelines expects the bare KFP IR PipelineSpec, so for TFX pipelines
+// the nested pipelineSpec is unwrapped. For non-TFX pipelines (e.g. kfp-sdk),
+// the compiled pipeline is returned unchanged.
+func unwrapTfxPipelineSpec(compiledPipeline json.RawMessage, frameworkName string) (json.RawMessage, error) {
+	if strings.ToLower(frameworkName) != "tfx" {
+		return compiledPipeline, nil
+	}
+
+	var wrapper map[string]json.RawMessage
+	if err := json.Unmarshal(compiledPipeline, &wrapper); err != nil {
```
chr12c, line 97: the fucniton never returns a error. I'm not sure if that's a mistake. At the very least an unmarshall error should return an error right?

provider-service/kfp/internal/provider/provider.go
```diff
@@ -81,6 +84,30 @@ func NewKfpProvider(config *config.Config) (*KfpProvider, error) {
 
 var _ resource.Provider = &KfpProvider{}
 
+// unwrapTfxPipelineSpec extracts the pipelineSpec from a TFX compiled pipeline.
+// TFX compiles pipelines into a Vertex PipelineJob wrapper
+//
+// Kubeflow Pipelines expects the bare KFP IR PipelineSpec, so for TFX pipelines
```
chr12c: I can't remember exactly but I believe unwrapping pipelineSpec will throw away some other fields. But those other fields might be available through some other avenue.

provider-service/kfp/internal/provider/provider.go
```diff
@@ -81,6 +84,30 @@ func NewKfpProvider(config *config.Config) (*KfpProvider, error) {
...
+// the nested pipelineSpec is unwrapped. For non-TFX pipelines (e.g. kfp-sdk),
+// the compiled pipeline is returned unchanged.
+func unwrapTfxPipelineSpec(compiledPipeline json.RawMessage, frameworkName string) (json.RawMessage, error) {
+	if strings.ToLower(frameworkName) != "tfx" {
```
chr12c, line 92: kfp and tfx compiled pipelines both run through this funciton but the name of this function suggests it already knows it's dealing with tfx. Rename the function (normalizePiplineSpec?) and/or split into a pure unwrapping + gate.

TobyPinfold: renamed extractPipelineSpec

provider-service/kfp/internal/provider/provider.go
TobyPinfold: now returns the error

provider-service/kfp/internal/provider/provider.go
TobyPinfold: The fields that are stripped off the wrapper is metadata which is already sent to VAI in the api call itself. The pipeline doesn't need the wrapper.

peter-hazell (approved):

chr12c (approved):

PR #1277, opened 2026-06-23
fix(vai-provider): handle 2.0.0 and 2.1.0 pipeline schemas

TobyPinfold:
links to #651 

  Vertex AI's `PipelineJob.RuntimeConfig` binds parameters via different fields depending on `pipelineSpec.schemaVersion`:

  - `<= 2.0.0` — uses `RuntimeConfig.Parameters` (deprecated).
  - `>= 2.1.0` — uses `RuntimeConfig.ParameterValues`.

  The VAI provider only populated the deprecated `Parameters` field. When TFX (or any compiler) emitted a 2.1.0 IR, Vertex failed the job with `INTERNAL_ERROR` because the runtime config and the submitted `pipelineSpec` disagreed.
 
 ## Changes

  - Parse `pipelineSpec.schemaVersion` into a `*semver.Version` on `PipelineValues`; reject invalid versions up-front.
  - In `DefaultJobEnricher.Enrich`, when the resolved schema version is `>= 2.1.0`, promote `RuntimeConfig.Parameters` onto `RuntimeConfig.ParameterValues` and clear `Parameters`. String, int
  and double variants are covered.
  - `MkRunPipelineJob` / `MkRunSchedulePipelineJob` keep building the deprecated `Parameters` field; the enricher (which is the only place that knows the IR version) does the swap. This keeps
  the builder schema-agnostic.

### DIFF
--- a/provider-service/vai/internal/provider/job_builder.go
+++ b/provider-service/vai/internal/provider/job_builder.go
@@ -91,8 +91,6 @@ func (jb DefaultJobBuilder) MkRunSchedulePipelineJob(
 		}
 	}
 
-	// Populated on the deprecated Parameters field; the enricher promotes
-	// to ParameterValues when schemaVersion >= 2.1.0.
 	templateUri, err := util.PipelineUri(
 		rsd.PipelineName,
 		rsd.PipelineVersion,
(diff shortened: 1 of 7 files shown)


provider-service/vai/internal/provider/job_builder.go
```diff
@@ -91,9 +91,8 @@ func (jb DefaultJobBuilder) MkRunSchedulePipelineJob(
 		}
 	}
 
-	// Note: unable to migrate from `Parameters` to `ParameterValues` at this
-	// point as `PipelineJob.pipeline_spec.schema_version` used by TFX
-	// is 2.0.0 see deprecated comment.
+	// Populated on the deprecated Parameters field; the enricher promotes
+	// to ParameterValues when schemaVersion >= 2.1.0.
```
chr12c: Comment seems to be in the wrong place (including the old comment).

chr12c (approved):

peter-hazell (approved):

PR #1286, opened 2026-07-01
fix(tfx-shim): force-exit TFX executor to avoid pure-virtual shutdown crash

sky-danielsantos:
## Summary

Fix a TFX executor shutdown crash (`pure virtual method called`) on KFP v2 by shipping a `tfx_kfp_v2_shim` that force-exits the executor after it writes its output metadata, so the C++ destructor chain never runs. Also pins the quickstart + compiler to TFX **1.16.0** and works around a KFP v2 launcher bug that chokes on GCS directory markers.

## Problem

On KFP v2, each TFX component runs its executor as `python -m tfx.orchestration.kubeflow.v2.container.kubeflow_v2_run_executor`. During interpreter shutdown — **after** the executor has already done its work and written its output metadata — the process aborts with `pure virtual method called` from the protobuf / ml-metadata C++ destructor chain. In the Trainer/Pusher this fails the component (and the whole pipeline) even though the model was produced correctly.

A runtime import hook alone cannot fix this: the executor is the `-m` entry module (`__main__`), and `runpy` loads it via the loader's `get_code`, which bypasses the import hook's `exec_module`. So a runtime patch never fires for the entry module.

Separately, TF's legacy GCS filesystem writes **zero-byte directory-marker blobs** (names ending in `/`). The KFP v2 launcher downloads these as files, which then blocks `os.MkdirAll` for child paths in downstream components (see kubeflow/pipelines#13476).

## Solution

- **Force-exit as a build-time source edit (Patch 8):** `install_shim.patch_run_executor_source` inserts `os._exit(0)` into the installed `kubeflow_v2_run_executor.py` immediately after the output-metadata write, skipping the crashing destructor chain. This covers the `-m` entry module where the runtime hook cannot; a runtime wrapper is kept as a fallback for direct (non-`-m`) imports.
- **Fail loud on install:** `main()` treats a failed source patch (missing file/anchor — e.g. an unexpected TFX version) as fatal (logs an error, exits non-zero), so an image can never silently ship without the crash fix.
- **GCS directory-marker cleanup (Patch 7):** after each executor run, `translate_executor_output` is wrapped to delete zero-byte `/`-suffixed blobs under every output artifact URI, unblocking downstream `MkdirAll`. Patches 5 and 7 are workarounds for kubeflow/pipelines#13476 and become redundant once that is fixed upstream.
- **Pin TFX 1.16.0** in `compilers/tfx` and the quickstart, and add the `dill` dependency required for the quickstart to compile under 1.16.0.
- **Quickstart cleanup:** wire the shim into the TFX quickstart image and remove the redundant hand-rolled `_cleanup_gcs_markers` helper from `trainer.py` — Patch 7 already cleans markers under every artifact URI (including the trainer's model dir), so the sample pipeline no longer carries an infra workaround.

## Testing

Validated end-to-end on GKE / Vertex AI (`europe-west2`): the TFX penguin quickstart compiled and ran to `PIPELINE_STATE_SUCCEEDED` with `CsvExampleGen`, `Trainer`, and `Pusher` all SUCCEEDED (Trainer/Pusher are the stages the crash previously hit). The `Trainer/model` and `Pusher/pushed_model` SavedModels are non-empty (`saved_model.pb` + `variables/` + `fingerprint.pb`), and a full GCS scan of the run tree found **no** zero-byte markers inside `model/`, `pushed_model/`, or `variables/`. The original crash was also reproduced and the force-exit fix confirmed on EKS / S3.

- `patch_run_executor_source` unit tests: applies the edit, is idempotent, and returns `False` on missing file / missing anchor.
- Shim import-hook and patch tests (`provider-service/kfp/tfx-patch/tests/test_shim.py`).


compilers/tfx/pyproject.toml
```diff
@@ -13,7 +13,7 @@ dependencies = [
 [project.optional-dependencies]
 dev = [
     "pytest==8.4.2",
-    "tfx==1.17.3",
+    "tfx==1.16.0",
```
TobyPinfold: Any reason for the downgrade? Prior investigations I found supporting 1.17.3 as. a dev dependency had no impact on the pipeline compilations of quickstart for all lower version 1.15.1 >

sky-danielsantos: It was the version specified on the ticket. It's also the one in https://github.com/NBCUDTC/prs-ml-commons/blob/master/ml-training-base/Makefile. 

Slack Thread: https://sky.slack.com/archives/C028114EPFD/p1781775583503789 

I can try and validate if the 1.17.3 is compatible with beam 2.64.0, if that is the desirable way forward.

TobyPinfold: I think the compiler can remain at 1.17.3, the pipeline can stay as 1.16

provider-service/kfp/tfx-patch/tfx_kfp_v2_shim/_patches.py
```diff
@@ -226,6 +233,39 @@ def patch_experimental(mod: ModuleType) -> None:
...
 
+# ── Patch 8 ──────────────────────────────────────────────────────────────
+# Issue:   On container exit after a TFX component runs, the interpreter
+#          shutdown sequence triggers C++ destructors in protobuf/MLMD that
+#          crash with "pure virtual method called" (seen on TFX 1.17.3 / S3).
+# Solution: Wrap _run_executor to call os._exit(0) once it returns, bypassing
+#           the shutdown that provokes the crash. The metadata write is the
+#           final statement of _run_executor, so exiting after it returns is
+#           behaviourally identical to exiting immediately after the write.
+# (Ported from the 711 exploration branch, where it was "Patch 6".)
```
TobyPinfold:
```suggestion
```

docs-gen/includes/master/quickstart/pyproject.toml
```diff
@@ -3,18 +3,25 @@ name = "penguin_pipeline"
 version = "0.1.0"
 description = ""
 authors = []
-requires-python = "==3.10.12"
+requires-python = ">3.10.19,<3.11"
```
chr12c: Be more specific on the lower bound, i.e. at least use `>=`

provider-service/kfp/tfx-patch/README.md
chr12c: this is still showing the old prebuilt image flow.

provider-service/kfp/tfx-patch/tfx_kfp_v2_shim/_patches.py
```diff
@@ -226,6 +233,39 @@ def patch_experimental(mod: ModuleType) -> None:
...
+    and ``_run_executor`` is called directly.
+    """
+    original = mod._run_executor
+
+    @functools.wraps(original)
+    def _patched(*args, **kwargs):
+        result = original(*args, **kwargs)
+        log.info("Executor complete; forcing os._exit(0) to avoid shutdown crash")
+        os._exit(0)
+        return result
```
chr12c:
```suggestion
```

unreachable

docs-gen/includes/master/quickstart/Dockerfile
```diff
@@ -1,13 +1,24 @@
 ARG UV_VERSION
 FROM ghcr.io/astral-sh/uv:${UV_VERSION} AS uv
 
-FROM python:3.10.12
+FROM python:3.10-trixie
```
chr12c: Not a pinned version. Change to a pinned version.

compilers/tfx/pyproject.toml
sky-danielsantos: Done

provider-service/kfp/tfx-patch/tfx_kfp_v2_shim/_patches.py
sky-danielsantos: Done

docs-gen/includes/master/quickstart/pyproject.toml
sky-danielsantos: Done

provider-service/kfp/tfx-patch/README.md
sky-danielsantos: Done

provider-service/kfp/tfx-patch/tfx_kfp_v2_shim/_patches.py
sky-danielsantos: Done

docs-gen/includes/master/quickstart/Dockerfile
sky-danielsantos: Done

TobyPinfold (approved):

bcarter97 (approved):

provider-service/kfp/tfx-patch/tfx_kfp_v2_shim/install_shim.py
chr12c: this docstring still needs to be updated

chr12c (approved):

PR #1295, opened 2026-07-07
feat(kfp-provider): resolve default experiment in provider layer

### LINKED ISSUE #1294: Controller should not know about DefaultExperiment
The DefaultExperiment is configured on the controller, but the concept of a default experiment is only understood by kubeflow pipelines, and therefore only the KFP Provider needs to know it.

Move the DefaultExperiment and the logic that uses this value into the KFP Provider.

bcarter97:
- Closes #1294

## Tasks

- [x] Removed `DefaultExperiment` and logic from the controller
- [x] Moved `DefaultExperiment` and the fallback logic to the provider
- [x] Documentation updated


controllers/pipelines/internal/workflowfactory/run_workflow_factory.go
```diff
@@ -22,16 +22,9 @@ type RunDefinitionCreator struct {
...
-	if run.Spec.ExperimentName == "" {
-		experimentName = common.NamespacedName{
-			Name: rdc.Config.DefaultExperiment,
-		}
-	} else {
-		experimentName = common.NamespacedName{
-			Name:      run.Spec.ExperimentName,
-			Namespace: run.Namespace,
-		}
+	experimentName := common.NamespacedName{
```
chr12c: If the user does not specify experiment name, then this `NamespacedName` will only have a namespace, but no name. The resource workflow factory will fail when it attempts to marshall this run definition. Same applies to the runschedule workflow factory.

chr12c: 87ba0058831604278bfe2c683a3b706f579cc6bb

chr12c (approved):

TobyPinfold (approved):

PR #1296, opened 2026-07-07
feat(provider): add podTemplateAnnotations as a configurable setting on controller level

chr12c:
Makes the provider more flexible by allowing it to set pod annotations.


chr12c:
Closing this for now because https://github.com/sky-uk/kfp-operator/pull/1297 is more useful since it is on a per-provider basis.

PR #1310, opened 2026-07-10
feat(provider): add podTemplateEnv to spec

### LINKED ISSUE #1311: Allow configuring pod template environment variables on a per-provider basis

Currently the only way to configure a Provider's environment variables is through the controller's ConfigMap. However this applies to all Providers only, and there's no way to configure them per Provider.

Introduce a new field on the Provider spec that allows users to configure arbitrary environment variables. The preferred way is to use k8s conventions, i.e. use `k8s.io/api/core/v1.EnvVar` because it accepts `valueFrom` as well.

chr12c:
Closes #1311 

Type this to the standard k8s EnvVar, because it also accepts `valueFrom` semantics.


apis/pipelines/hub/provider_types.go
```diff
@@ -42,6 +43,11 @@ type ProviderSpec struct {
 	Parameters          map[string]*apiextensionsv1.JSON `json:"parameters,omitempty" yaml:"parameters,omitempty"`
 	Frameworks          []Framework                      `json:"frameworks,omitempty" yaml:"frameworks,omitempty"`
 	AllowedNamespaces   []string                         `json:"allowedNamespaces,omitempty" yaml:"allowedNamespaces,omitempty"`
+	// +kubebuilder:validation:Optional
+	// Environment variables applied to the generated provider-service container,
+	// merged over the operator's defaultProviderValues env; per-provider values
```
bcarter97, line 48: This could be a problem if a derived environment variable used for the providers e.g. `PROVIDERNAME` was set... unlikely but possible

bcarter97 (commented):
Docs need updating in https://github.com/sky-uk/kfp-operator/blob/master/docs-gen/content/docs/reference/resources/provider.md

apis/pipelines/hub/provider_types.go
chr12c: That's correct. In fact there is a risk of all the environment variables of being set, and this includes existing variables. An example would be any variable that is in the form `PARAMETERS_*`. This should be a separate issue since it's already an existing thing.

chr12c:
> Docs need updating in https://github.com/sky-uk/kfp-operator/blob/master/docs-gen/content/docs/reference/resources/provider.md

b4d5c7d5f922cbbe038bbce44a71d9c2905bcede

bcarter97 (approved):

TobyPinfold (approved):

PR #1314, opened 2026-07-13
refactor(vai-provider): move JobBuilder and FileHandler interfaces to consumer

chr12c:
Make the aforementioned provider interfaces consumer-driven, which follows golang best practices.

### DIFF
--- a/provider-service/vai/internal/provider/provider.go
+++ b/provider-service/vai/internal/provider/provider.go
@@ -60,12 +60,29 @@ type jobEnricher interface {
 	) (*aiplatformpb.PipelineJob, error)
 }
 
+type jobBuilder interface {
+	MkRunPipelineJob(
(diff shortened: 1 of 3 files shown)


provider-service/vai/internal/provider/provider.go
```diff
@@ -47,19 +47,42 @@ type scheduleClient interface {
 	) (*aiplatformpb.Schedule, error)
 }
 
+type fileHandler interface {
```
bcarter97, line 50: why is it camelCase in this file but not the other one? some convention?

chr12c: you mean the `FileHandler` it replaced? It's because this interface is private and has no reason to be exported.

TobyPinfold (approved):

bcarter97 (approved):

PR #1315, opened 2026-07-13
feat(providers): isolate argo-workflows at provider level 

TobyPinfold:
Closes #428 

Isolates argo-workflows created by the operator to provider namespaces. This is to ensure 2 different providers are not resource starved if one is creating a surge of workflow operations at any given time. With this change the providers resource allocation is isolated from other providers and its resource ceiling is clearly defined.

## Tasks

- [x] Documentation updated


controllers/pipelines/internal/workflowfactory/resource_workflow_factory.go
```diff
@@ -57,16 +57,18 @@ const (
 // Template name methods - create and update operations use suffixed templates to differentiate
 // between resource types (e.g., "compiled" for pipelines, "simple" for basic resources).
 // Delete operations use a single shared template since deletion is generic across all resource types.
+// Templates are provisioned per-provider-namespace by the kfp-provider-workflows Helm chart under
+// fixed names, so the operator references them by their static names.
```
chr12c, line 61: This comment implies it was previously not static, and now it is somehow static. I don't see how that's the case: it's just as static/non-static as it was before the template prefix was removed, e.g. right now the create/update templates have a suffix in the name, which is the same as before.

helm/kfp-operator/charts/kfp-provider-workflows/Chart.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v2
+name: kfp-provider-workflows
```
chr12c, line 2: This helm chart pushes the idea that the provider resource is not self sufficient even more so than before. A prerequisite of the resource is to have the various workflow templates deployed and the rbac. We can't just rely on the provider resource being the source of truth. The provider should also come with the templates + rbac, along with the existing Deployment. Requiring a helm chart to drive this feels like an anti-pattern.

controllers/pipelines/internal/workflowfactory/resource_workflow_factory.go
TobyPinfold: before it was prefixed by the kfp-operator installation namespace, now they are explicitly statically named.

helm/kfp-operator/charts/kfp-provider-workflows/Chart.yaml
TobyPinfold: imo the helm chart was more for convenience, I don't see this as being any different to how the templates where being applied via the operators helm template. I equally don't think they should be created by the operator.

helm/kfp-operator/charts/kfp-provider-workflows/values.yaml
```diff
@@ -0,0 +1,60 @@
...
+# Install one release per provider namespace.
+
+# Namespace the WorkflowTemplates, RBAC and Provider resource are created in.
+# This must be the namespace the Provider runs in. Defaults to the release
+# namespace when left empty.
+namespace: ""
+
+# The Argo execution configuration. Mirrors manager.argo in the kfp-operator
+# chart so provider-namespace templates stay consistent with the operator.
+manager:
```
chr12c: "manager" is a misleading wrapper key. There's nothing here that has anything to do with the kfp operator controller

chr12c (approved):

helm/kfp-operator/values.yaml
peter-hazell: Is any of this still needed now that the argo config is in a different chart?

docs-gen/includes/master/reference/controller_manager_config.yaml
```diff
@@ -4,8 +4,9 @@ system:
     id: kfp-operator-lock
 spec:
   multiversion: true
+  # Deprecated: resource-management workflows now run in each Provider's own
+  # namespace. Retained only as the namespace for the run-completion EventBus.
```
peter-hazell: wonder if we should rename this to something more suitable then if it's not used for workflows any more and we still need it for the EventBus?

helm/kfp-operator/values.yaml
TobyPinfold: [3d24090](https://github.com/sky-uk/kfp-operator/pull/1315/commits/3d240906229762e01befcbddd8a76385e5cbeeeb)

docs-gen/includes/master/reference/controller_manager_config.yaml
TobyPinfold: [d0e79db](https://github.com/sky-uk/kfp-operator/pull/1315/commits/d0e79dbf947a6abcd9b05a757500eb5324400d94)

helm/kfp-operator/charts/kfp-provider-workflows/values.yaml
TobyPinfold: [1b0832b](https://github.com/sky-uk/kfp-operator/pull/1315/commits/1b0832b1a61f651d9b4e75f41ad867f62164c7f0)

peter-hazell (approved):

PR #1334, opened 2026-07-18
feat(kfp provider): scope kubeflow pipelines api server requests to provider namespace in multi user mode

### LINKED ISSUE #1335: Support multi user isolation in KFP provider

Kubeflow Pipelines has a [multi user isolation feature](https://www.kubeflow.org/docs/components/pipelines/operator-guides/multi-user/). In order to support this mode, KFP API server experiment service calls need to be namespaced.

The assumption is multi user isolation will be on the provider namespace level, therefore calls to the experiment service need to be directed to the provider namespace.
Single user mode still needs to work, i.e. no namespace being sent to the experiment service in this mode.

chr12c:
Closes #1335

Support Kubeflow Pipelines multi user mode and maintain backwards compatibility with single user mode.

### DIFF
--- a/provider-service/kfp/internal/provider/experiment_service.go
+++ b/provider-service/kfp/internal/provider/experiment_service.go
@@ -36,8 +36,7 @@ type DefaultExperimentService struct {
 // The namespace sent on experiment requests is fixed for the lifetime of the
 // service. KFP multi-user mode requires every experiment request to be scoped
 // to a namespace, so requests are pinned to providerNamespace; single-user mode
-// requires the namespace to be empty. This is resolved once here rather than per
-// request because it never varies with the incoming request.
+// requires the namespace to be empty.
 func NewExperimentService(
 	conn *grpc.ClientConn,
 	multiUserMode bool,
--- a/provider-service/kfp/internal/provider/experiment_service.go
+++ b/provider-service/kfp/internal/provider/experiment_service.go
@@ -106,6 +106,10 @@ func (es *DefaultExperimentService) ExperimentIdByDisplayName(
 	ctx context.Context,
 	experiment common.NamespacedName,
 ) (string, error) {
+	if experiment.Namespace == "" {
+		experiment.Namespace = es.requestNamespace
+	}
+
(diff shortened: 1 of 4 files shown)


sky-danielsantos (approved):

bcarter97 (approved):
Pipelines aren't NS scoped. Assume this is intentional and they will go into the global NS?

chr12c:
> Pipelines aren't NS scoped. Assume this is intentional and they will go into the global NS?

the KFP provider does not submit a namespace when communicating with Kubeflow Pipelines API Server for pipelines, so they do indeed go to the "global" namespace. Kubeflow does support namespaced calls, but we just don't do it here

provider-service/kfp/internal/provider/experiment_service.go
```diff
@@ -25,20 +25,39 @@ type ExperimentService interface {
...
+	if multiUserMode {
+		requestNamespace = providerNamespace
+	}
+
 	return &DefaultExperimentService{
-		client: go_client.NewExperimentServiceClient(conn),
+		client:           go_client.NewExperimentServiceClient(conn),
+		multiUserMode:    multiUserMode,
+		requestNamespace: requestNamespace,
 	}, nil
```
TobyPinfold, line 66:
```suggestion
	if multiUserMode && providerNamespace == "" {
		return nil, errors.new("Multi-user mode requires a provider namespace be provided")
	}

return &DefaultExperimentService{
		client:           go_client.NewExperimentServiceClient(conn),
		multiUserMode:    multiUserMode,
		providerNamespace: providerNamespace,
}, nil
```

So it fails fast instead of default to "" for multi user mode

chr12c: I understand the gist of your suggestion and implemented it correctly here: 8c0402c1a44baa18375e95c060c533f016c83e47

Did not use your suggestion verbatim as it's actually incorrect. It's also worth noting that none of this is reachable code anyway since k8s API guarantees the presence of the provider namespace. But I'll add it for completeness.

TobyPinfold (approved):

PR #1346, opened 2026-07-28
feat(kfp provider): support bearer token auth multi user

### LINKED ISSUE #1350: Support multi user Kubeflow Pipelines authentication to ml-pipeline

Support multi user Kubeflow Pipelines authentication to ml-pipeline. Currently there is no auth plumbing in the KFP provider itself, so the aim is to implement this plumbing within the gRPC and REST clients.

read and send the mounted token to the KFP API server on each call.

chr12c:
Closes #1350

Expects a configured pod volume and volume mount for a token specifically for kubeflow pipelines, and is only used in multi user mode.


provider-service/base/pkg/auth/rest.go
```diff
@@ -0,0 +1,21 @@
+package auth
+
+import (
+	"github.com/go-openapi/runtime"
+	"github.com/go-openapi/strfmt"
+	"golang.org/x/oauth2"
+)
+
+// BearerAuthInfoWriter returns a runtime.ClientAuthInfoWriter that adds an
+// "Authorization: Bearer <token>" header to go-openapi REST requests.
```
bcarter97, line 10: Think this is pretty self-evident

```suggestion
```

provider-service/base/pkg/auth/rest.go
bcarter97, line 1: If this is a specific openapi impl, should we have a sub-package for http and put this in open_api.go?

provider-service/kfp/internal/client/kfp_api.go
```diff
@@ -69,18 +73,75 @@ func (gka *GrpcKfpApi) GetResourceReferences(ctx context.Context, runId string)
 	return resourceReferences, nil
 }
 
-func CreateKfpApi(ctx context.Context, config config.Config) (*GrpcKfpApi, error) {
+// dialer matches grpc.NewClient so it can be swapped out in tests.
```
bcarter97: Not sure I understand why this is needed

peter-hazell (approved):

provider-service/kfp/internal/client/kfp_api.go
chr12c: I must've added it thinking it was required in order to test the dial options in the yes-tokensource/no-tokensource cases.

Removed abstraction: f3caaabb1cd6e31ecfe73dc202740401617f184a

provider-service/base/pkg/auth/rest.go
chr12c: It's a doc comment on an exported function, so by Go convention this is fine. Deleting it doesn't just remove a comment, but causes the exported API to be undocumented.

provider-service/base/pkg/auth/rest.go
chr12c: the auth package is split symmetrically by transport (grpc and rest), each with only a small amount of code tied to one client lib each. A subpackage just for the REST side would break that symmetry for little gain. I think we should keep it flat for now. Happy to revisit if we add a 2nd HTTP/REST client that needs its own auth writer.

bcarter97 (approved):

PR #1349, opened 2026-07-29
feat:(provider): support Volumes and VolumeMounts

### LINKED ISSUE #1347: Allow configuring Provider volumes and volume mounts

It would be useful to be able to configure the Provider's volumes and volume mounts, for each Provider individually. There can be many use cases, but the one that comes to mind is being able to facilitate bearer token authentication.

Make Provider crd open for configuring volumes and volume mounts

bcarter97:
Closes #1347 

## Tasks

- [ ] add `podTemplateVolumes` and `podTemplateVolumeMounts` to the Provider CRD
- [x] Documentation updated


controllers/pipelines/provider_deployment_manager.go
```diff
@@ -161,11 +161,14 @@ func populateServiceContainer(serviceContainerName string, podTemplate corev1.Po
...
 	})
 
+	podTemplate.Spec.Volumes = append(podTemplate.Spec.Volumes, provider.Spec.PodTemplateVolumes...)
+
 	podTemplate.Spec.Containers = lo.Map(podTemplate.Spec.Containers, func(c corev1.Container, _ int) corev1.Container {
 		if c.Name == serviceContainerName {
 			c.Image = provider.Spec.ServiceImage
 			c.Env = append(c.Env, envVars...)
 			c.Env = mergeEnvByName(c.Env, provider.Spec.PodTemplateEnv)
+			c.VolumeMounts = append(c.VolumeMounts, provider.Spec.PodTemplateVolumeMounts...)
```
chr12c: the appends stack the provider values on top of the operator defaults. So if a provider declares a volume with the same name as a default volume, then the append will produce 2 entires with same name. The API server will reject the whole Deployment. We should have a mergeVolumesByName and mergeVolumeMountsByPath which work in a similar way to the existing mergeEnvByName.

bcarter97: https://github.com/sky-uk/kfp-operator/pull/1349/commits/2c419bc35089aa9fb586dddb3e5d153f186cfe35

apis/pipelines/hub/provider_types.go
```diff
@@ -48,6 +48,14 @@ type ProviderSpec struct {
...
 	// take precedence on name collision.
 	PodTemplateEnv []corev1.EnvVar `json:"podTemplateEnv,omitempty" yaml:"podTemplateEnv,omitempty"`
+	// +kubebuilder:validation:Optional
+	// Volumes added to the generated provider-service pod, appended to the
+	// operator's defaultProviderValues volumes.
+	PodTemplateVolumes []corev1.Volume `json:"podTemplateVolumes,omitempty" yaml:"podTemplateVolumes,omitempty"`
+	// +kubebuilder:validation:Optional
+	// VolumeMounts added to the generated provider-service container, paired
+	// with PodTemplateVolumes by name.
+	PodTemplateVolumeMounts []corev1.VolumeMount `json:"podTemplateVolumeMounts,omitempty" yaml:"podTemplateVolumeMounts,omitempty"`
```
chr12c, line 60: these comments need to be updated to reflect the fact that collisions are handled (see podTemplateEnv)

apis/pipelines/hub/provider_webhook.go
```diff
@@ -73,6 +73,20 @@ func (v *ProviderValidator) validate(
...
+	for i, volumeMount := range provider.Spec.PodTemplateVolumeMounts {
+		if _, ok := volumeNames[volumeMount.Name]; !ok {
+			errs = append(errs, field.Invalid(
+				field.NewPath("spec", "podTemplateVolumeMounts").Index(i).Child("name"),
+				volumeMount.Name,
+				"volume mount does not reference a volume declared in podTemplateVolumes",
+			))
+		}
+	}
+
```
chr12c: the webhook only checks mount names against the provider's podTemplateVolumes, but the running pod also gets volumes merged from the operator's default provider values. E.g. the operator can declare a global volume, the provider can declare the mount, but it gets rejected at admission even though the pod will be fine. The check should be dropped and just fallback to the reconcile to handle it.

apis/pipelines/hub/provider_types.go
bcarter97: https://github.com/sky-uk/kfp-operator/pull/1349/commits/4145f26b0e59a5c8604daf057835b5dc45693424

apis/pipelines/hub/provider_webhook.go
bcarter97: https://github.com/sky-uk/kfp-operator/pull/1349/commits/26c1e14a2811cedc4cca566d1bbc7a04e271122a

controllers/pipelines/provider_deployment_manager.go
```diff
@@ -161,42 +161,58 @@ func populateServiceContainer(serviceContainerName string, podTemplate corev1.Po
...
-// overrides replace base entries with the same name (preserving position) and are
-// otherwise appended, so per-provider env takes precedence on name collision.
-func mergeEnvByName(base []corev1.EnvVar, overrides []corev1.EnvVar) []corev1.EnvVar {
-	merged := make([]corev1.EnvVar, len(base))
+// mergeByKey merges overrides into base, keyed on the value returned by key.
+// Entries in overrides replace base entries with the same key (preserving position)
+// and are otherwise appended, so overrides take precedence on key collision.
+func mergeByKey[T any, K comparable](base []T, overrides []T, key func(T) K) []T {
+	merged := make([]T, len(base))
 	copy(merged, base)
```
chr12c:
```suggestion
	merged := slices.Clone(base)
```

chr12c (approved):

TobyPinfold (approved):

PR #1370, opened 2026-08-12
fix(informer-cache): only cache argo-workflows created by the operator.

### LINKED ISSUE #1369: Cache memory is being wasted on unrelated argo-workflows.
### Problem
  The controller-manager watches Argo `Workflow` objects with no label selector on the manager cache. controller-runtime therefore lists/watches and holds **every** Workflow in the cluster in
  memory not just the ones the operator created.

  In clusters where other systems also create Argo Workflows, the operator's cache can be dominated by unrelated Workflows it will never reconcile. This inflates the controller-manager's
  memory footprint (in an environment with many workflows a majority cached Workflows may be unrelated to the operator), which can push the pod against its memory limit and trigger OOMs

  ### Cause
  - `main.go` — the manager `Cache` has no per-object selector, so the informer caches Workflows cluster-wide.
  - The owner-label check in the map func runs *after* caching, so it only suppresses reconciles, not memory.

  ### Impact
  Restricts the cache to operator-owned Workflows only, removing the bulk of the manager's memory footprint in clusters with unrelated Argo Workflows.

TobyPinfold:
Closes #1369 

This adds a filter to the informer cache to only store workflows that are created by the operator itself. This prevents memory space being wasted on argo workflows that are irrelevant to the operator.

### DIFF
--- a/main.go
+++ b/main.go
@@ -91,13 +91,18 @@ func main() {
 	opts.BindFlags(flag.CommandLine)
 	flag.Parse()
 
-	var err error
 	ctrlConfig := config.OperatorConfig{}
 
+	// Create a workflow requirement such that only workflows that are owned by the operators api group are cached.
 	ownerKindLabelKey := apis.Group + "/owner.kind"
-	workflowOwnerReq, _ := labels.NewRequirement(
+	workflowOwnerReq, err := labels.NewRequirement(
 		ownerKindLabelKey, selection.Exists, nil,
 	)
+	if err != nil {
+		setupLog.Error(err, "unable to create workflow owner requirement")
+		os.Exit(1)
+	}
+
 	options := ctrl.Options{
 		Scheme:                 scheme,
 		HealthProbeBindAddress: ":8081",
--- a/main.go
+++ b/main.go
@@ -22,11 +22,17 @@ import (
 	"net/http"
 	"os"
 
+	argo "github.com/argoproj/argo-workflows/v3/pkg/apis/workflow/v1alpha1"
 	"github.com/samber/lo"
+	"github.com/sky-uk/kfp-operator/apis"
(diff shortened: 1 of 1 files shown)


peter-hazell (approved):

main.go
```diff
@@ -87,7 +93,22 @@ func main() {
...
+	workflowOwnerReq, _ := labels.NewRequirement(
+		ownerKindLabelKey, selection.Exists, nil,
+	)
+	options := ctrl.Options{
+		Scheme:                 scheme,
+		HealthProbeBindAddress: ":8081",
+		Cache: cache.Options{
+			ByObject: map[client.Object]cache.ByObject{
+				&argo.Workflow{}: {
+					Label: labels.NewSelector().Add(*workflowOwnerReq),
```
chr12c, line 112: Not obvious why this is needed so a short comment would help

main.go
```diff
@@ -87,7 +93,22 @@ func main() {
 
 	var err error
 	ctrlConfig := config.OperatorConfig{}
-	options := ctrl.Options{Scheme: scheme, HealthProbeBindAddress: ":8081"}
+
+	ownerKindLabelKey := apis.Group + "/owner.kind"
+	workflowOwnerReq, _ := labels.NewRequirement(
```
chr12c: error is discarded

TobyPinfold: updated in [22bed50](https://github.com/sky-uk/kfp-operator/pull/1370/commits/22bed503d41ab5b4882e575e431ab5166a941f16)

main.go
TobyPinfold: Added comment to [22bed50](https://github.com/sky-uk/kfp-operator/pull/1370/commits/22bed503d41ab5b4882e575e431ab5166a941f16)

chr12c (approved):

